PR #1, opened 2022-10-09
Add code examples for types and possible values

ggawryal:
Add a few correct and incorrect examples of programs in complied language for types and allowed values of type literals.


ggawryal:
Closed as we discussed that examples should be put only into language specification doc.

PR #11, opened 2022-10-18
Parser grammar and analysis template

ggawryal:


app/src/main/kotlin/compiler/parser/grammar/AutomatonGrammar.kt
```diff
@@ -0,0 +1,11 @@
+package compiler.parser.grammar
+import compiler.common.dfa.state_dfa.TransparentDfa
+
+class AutomatonGrammar<S : Comparable<S>> {
+    fun getState(): S {
```
JHodorHodor, line 5: I think that this was supposed to be `getStartState`.

app/src/main/kotlin/compiler/common/dfa/DfaWalk.kt
```diff
@@ -0,0 +1,13 @@
+package compiler.common.dfa
+
+interface DfaWalk<A, R> {
+    fun getAcceptingStateTypeOrNull(): R?
+
+    fun isAccepting(): Boolean {
```
tomekb234, line 6: One shouldn't be able to override this method. We can replace it with an extension method.

app/src/main/kotlin/compiler/common/dfa/state_dfa/DfaState.kt
```diff
@@ -0,0 +1,6 @@
+package compiler.common.dfa.state_dfa
+
+interface DfaState<A, R> {
+    val result: R?
+    fun getPossibleSteps(): Map<A, DfaState<A, R>>
```
tomekb234, line 5: Seems somewhat inconsistent that `result` is a property and `getPossibleSteps` is a method. (Maybe we could also replace any `getX` with `x` elsewhere in this project.)

app/src/main/kotlin/compiler/common/dfa/state_dfa/SingleAcceptingState.kt
```diff
@@ -0,0 +1,5 @@
+package compiler.common.dfa.state_dfa
+
+enum class SingleAcceptingState {
```
tomekb234, line 3: We can just use `Unit` instead.

app/src/main/kotlin/compiler/common/dfa/state_dfa/TransparentDfa.kt
```diff
@@ -0,0 +1,10 @@
+package compiler.common.dfa.state_dfa
+
+import compiler.common.dfa.Dfa
+
+interface TransparentDfa<A, R> : Dfa<A, R> {
```
tomekb234, line 5: Maybe `AbstractDfa` and `Dfa` instead of `Dfa` and `TransparentDfa`? (I suppose we'll use the second one more often.)

app/src/main/kotlin/compiler/common/dfa/state_dfa/DfaState.kt
ggawryal: Agree, changed most getters to properties, at least in `parser` module

app/src/main/kotlin/compiler/common/dfa/state_dfa/TransparentDfa.kt
ggawryal: This is better, but it still doesn't tell the difference between those two interfaces. We can rename it later if we came up with even better name.

tomekb234 (approved):

PR #13, opened 2022-10-21
Tests for derivatives of star, union, concat

woocash2:
Provided tests which we lacked before because I didn't know if I was going to perform these tests in isolation from RegexFactory.
Finally decided not to isolate them.


ggawryal (approved):
LGTM, very exhaustive tests. Maybe it could be useful to add a very simple test for concat, e.g. reg=`ab`, because the regexes we have are not the most trivial ones (e.g. for star we're testing in fact both derivatives of star and concat), but since it wouldn't change the code coverage, it's up to you.

PR #14, opened 2022-10-21
Add initial Abstract Syntax Tree

tomekb234:


app/src/main/kotlin/compiler/ast/StatementAst.kt
```diff
@@ -0,0 +1,34 @@
...
+    data class Loop(
+        val condition: ExpressionAst,
+        val action: BlockAst
+    ) : StatementAst()
+
+    object LoopBreak : StatementAst()
+
+    object LoopContinuation : StatementAst()
+
+    data class FunctionReturn(val value: ExpressionAst?) : StatementAst()
```
ggawryal, line 33: Shouldn't we always return some value (for `Unit` functions `zakończ` will be translated to `zwróć Unit`)?

app/src/main/kotlin/compiler/ast/ExpressionAst.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    data class UnaryOperation(
+        val kind: Kind,
+        val operand: ExpressionAst
+    ) : ExpressionAst() {
+        enum class Kind {
+            NOT,
+            NEG,
+            BIT_NOT,
+        }
```
ggawryal, line 28: Just to confirm: we don't want to have an unary `plus`, right?

app/src/main/kotlin/compiler/ast/StatementAst.kt
```diff
@@ -0,0 +1,34 @@
+package compiler.ast
+
+typealias BlockAst = List<StatementAst>
+
+sealed class StatementAst {
```
ggawryal, line 5: Sometimens `Comments` are also included in AST, do we want to add them?

ggawryal (approved):
Dzięki, wygląda dobrze. Zostawiłem kilka pytań bardziej odnośnie języka niż kodu.

app/src/main/kotlin/compiler/ast/StatementAst.kt
tomekb234: I think this translation should be performed by the type checker, and not by parser. The AST should reflect the code as seen by the programmer. For the same reason the field `returnType` in `FunctionAst` is optional.

app/src/main/kotlin/compiler/ast/ExpressionAst.kt
tomekb234: We can add it, but I suppose the language won't become unusable without it.

app/src/main/kotlin/compiler/ast/StatementAst.kt
tomekb234: The compiler probably won't make any use of them, so I don't see a point.

app/src/main/kotlin/compiler/ast/ExpressionAst.kt
tomekb234: Added, and made the operation names more readable.

PR #18, opened 2022-10-22
Implement changes to Regex and RegexDfa

rkilar:
Added the `first` function to Regex as well as a `hashCode` function consistent with `equals`. Implemented RegexDfaState and added some tests for it. Moved test files for Regex and Dfa to packages that correspond to main source code.


app/src/main/kotlin/compiler/common/dfa/RegexDfaState.kt
```diff
@@ -1,11 +1,12 @@
...
-        get() = TODO("Not yet implemented")
+data class RegexDfaState<A : Comparable<A>>(private val regex: Regex<A>) : DfaState<A, Unit> {
+    override val result: Unit?
+        get() = if (regex.containsEpsilon()) Unit else null
 
-    override val possibleSteps: Map<A, DfaState<A, R>>
-        get() = TODO("Not yet implemented")
+    override val possibleSteps: Map<A, DfaState<A, Unit>>
+        get() = regex.first().associateWith { RegexDfaState(regex.derivative(it)) }
 }
```
bebidek, line 12: `RegexDfaState` should also have `equals` and `hashCode`.

rkilar: It is defined as a `data class`. Kotlin automatically generates `equals` and `hashCode` based on properties defined in the constructor, which in this case is just `regex`.

app/src/test/kotlin/compiler/common/dfa/RegexDfaTest.kt
```diff
@@ -0,0 +1,242 @@
...
+        for (word in words) {
+            val walk = regexDfa.newWalk()
+            word.forEach { walk.step(it) }
+
+            val shouldStateBeAccepting = word.lastOrNull() == 'b' && word.all { setOf('a', 'b').contains(it) }
+            assertEquals(shouldStateBeAccepting, walk.isAccepting(), "Is walk over $word accepted?")
+        }
+    }
+
+    @Test fun `test possible steps, results and predecessors on an example regex`() {
```
ggawryal, line 95: Can we move the regex outside that function and split this testcase into testing `possibleSteps`, `predecessors` and `results` separately?

app/src/test/kotlin/compiler/common/dfa/RegexDfaTest.kt
```diff
@@ -0,0 +1,242 @@
...
+
+        assertEquals(regexDfa.getAcceptingStates(), setOf(stateEPS))
+        assertNull(stateBC.result)
+        assertNull(stateC.result)
+        assertEquals(stateEPS.result, Unit)
+        assertNull(stateABStarC.result)
+    }
+
+    @Test fun `test possible steps, results and predecessors on example regex 2`() {
+        val regex: Regex<Char> = RegexParser.parseStringToRegex("([ab][cd])*a*")
```
ggawryal, line 184: Same here

app/src/main/kotlin/compiler/common/regex/Regex.kt
```diff
@@ -59,13 +87,30 @@ sealed class Regex<A : Comparable<A>> : Comparable<Regex<A>> {
         override fun equals(other: Any?): Boolean {
             return other is Epsilon<*>
         }
+
+        private companion object HashObject
+        override fun hashCode(): Int {
+            // Not sure if this is a good solution
```
ggawryal, line 93: This comment too

ggawryal (changes requested):
Thanks - looks nice. Left some comments, especially I wonder if the default implementation of `isDead` for walks is good enough.

app/src/main/kotlin/compiler/common/dfa/state_dfa/Dfa.kt
```diff
@@ -14,7 +14,14 @@ interface Dfa<A, R> : AbstractDfa<A, R> {
             }
 
             override fun isDead(): Boolean {
-                return currentState == null || (currentState?.result == null && currentState?.possibleSteps?.isEmpty() == true)
+                var canAccept = false
+                currentState?.let {
+                    dfs(it) { visitedState ->
+                        println("$visitedState ${visitedState.result != null}")
```
ggawryal, line 20: I guess we can remove this

ggawryal (approved):
LGTM!

PR #19, opened 2022-10-23
Automaton grammar generation

bebidek:
This is draft pull request because it depends on regex-additions PR. When regex-additions are merged, I will change it.

### DIFF
--- a/app/src/main/kotlin/compiler/common/dfa/CompositeDfa.kt
+++ b/app/src/main/kotlin/compiler/common/dfa/CompositeDfa.kt
@@ -1,0 +1,32 @@
+package compiler.common.dfa
+
+import compiler.common.dfa.state_dfa.Dfa
+import compiler.common.dfa.state_dfa.DfaState
+
+class CompositeDfa<S : Comparable<S>, R, IR> (private val components: List<Pair<Dfa<S, IR>, R>>) : Dfa<S, R> {
+    class MultipleAcceptingComponentsException : Exception()
+
+    inner class State(private val componentStates: List<DfaState<S, IR>?>) : DfaState<S, R> {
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/compiler/common/dfa/CompositeDfa.kt
```diff
@@ -0,0 +1,32 @@
+package compiler.common.dfa
+
+import compiler.common.dfa.state_dfa.Dfa
+import compiler.common.dfa.state_dfa.DfaState
+
+class CompositeDfa<S : Comparable<S>, R, IR> (private val components: List<Pair<Dfa<S, IR>, R>>) : Dfa<S, R> {
```
ggawryal, line 6: Currently we ignore the result type for composed dfas - I guess we can change `IR` to `Unit` (suggested) or modify `components` type to `List<Triple<Dfa<S, IR>, IR, R>>` (probably we don't need such functionality)

ggawryal (changes requested):
Great contribution, left 1 comment. Regex additions have been merged, so please merge with current `master`.

app/src/main/kotlin/compiler/common/dfa/CompositeDfa.kt
bebidek: I changed `IR` to `*` (so it's basically your first suggestion but hardcoding `Unit` when it's not necessary seems little weird to me).

bebidek:
I changed `AutomatonGrammar.startState` to `start` because grammar doesn't have states (maybe, if `start` is too vague, it could be `startSymbol`).

ggawryal (approved):
LGTM

PR #25, opened 2022-10-23
Implement follow

kamgal00:


app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
```diff
@@ -12,6 +18,89 @@ class GrammarAnalysis<S : Comparable<S>> {
...
+                val visited: MutableSet<DfaState<S, Production<S>>> = acceptingStates.toHashSet()
+                val queue: Queue<DfaState<S, Production<S>>> = LinkedList()
+                queue.addAll(acceptingStates)
+
+                while (!queue.isEmpty()) {
+                    val current = queue.remove()
+                    for ((symbol, predecessors) in dfa.getPredecessors(current)) {
+                        if (predecessors.isNotEmpty()) symbolsToPossibleEndSymbolsMap.getOrPut(lhs) { HashSet() }.add(symbol)
+                        if (symbol in nullable) {
+                            for (prev in predecessors) {
```
ggawryal, line 75: We can reduce nesting `if`s and `for`s:
```
predecessors.filter { it !in visited }.forEach {
    visited.add(it)
    queue.add(it)
}
```

app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
```diff
@@ -12,6 +18,89 @@ class GrammarAnalysis<S : Comparable<S>> {
...
+            while (wasResultUpdated) {
+                wasResultUpdated = false
+                for ((lhs, dfa) in grammar.productions) {
+                    val statesFirst = statesFirstMap.getOrPut(lhs) { HashMap() }
+                    val acceptingStates = dfa.getAcceptingStates()
+                    val visited: MutableSet<DfaState<S, Production<S>>> = acceptingStates.toHashSet()
+                    val queue: Queue<DfaState<S, Production<S>>> = LinkedList()
+                    queue.addAll(acceptingStates)
+
+                    while (!queue.isEmpty()) {
```
ggawryal, line 43: Similar bfs logic for accepting states backwards is used here twice and also for nullable, maybe we can create abstract bfs with action lambda as argument and use it to avoid repetition.

app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
```diff
@@ -12,6 +18,89 @@ class GrammarAnalysis<S : Comparable<S>> {
     }
 
     fun computeFollow(grammar: AutomatonGrammar<S>, nullable: Set<S>, first: Map<S, Set<S>>): Map<S, Set<S>> {
-        return TODO()
+
+        val result: MutableMap<S, MutableSet<S>> = HashMap()
+
+        var wasResultUpdated = true
+        fun update(setToUpdate: MutableSet<S>, secondSet: Set<S>) {
```
ggawryal, line 25: Maybe rename function to `updateResult` and remove `setToUpdate` argument, since we'll be always updating result after applying modification in 4th comment.

ggawryal (approved):
LGTM, left some comments to improve code quality. Some refactors may be easier to do later after we merge all three function, so we might discuss it tomorrow and don't care about them right now.

app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
kamgal00: Maybe this one will be easier to do after merging

app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
```diff
@@ -12,6 +18,89 @@ class GrammarAnalysis<S : Comparable<S>> {
...
+                    val acceptingStates = dfa.getAcceptingStates()
+                    val visited: MutableSet<DfaState<S, Production<S>>> = acceptingStates.toHashSet()
+                    val queue: Queue<DfaState<S, Production<S>>> = LinkedList()
+                    queue.addAll(acceptingStates)
+
+                    while (!queue.isEmpty()) {
+                        val current = queue.remove()
+                        for ((symbol, predecessors) in dfa.getPredecessors(current))
+                            for (prev in predecessors) {
+                                update(statesFirst.getOrPut(prev) { HashSet() }, first.getOrDefault(symbol, HashSet()))
```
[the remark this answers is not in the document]
kamgal00: Are you sure not using update here would be safe? I think it is possible that in some iterations only `statesFirst` updates while `result` does not, but in the next iteration changes made in `statesFirst` affect `result`

app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
kamgal00: I'm not sure, I also use it in line 48 to update statesFirst

ggawryal (approved):
It's good to be merged

app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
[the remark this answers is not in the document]
ggawryal: You're right. I thought that if we updated `statesFirst[prev]`, then later in the bfs in line 49  iteration we'll update result with `statesFirst[current]` for `current = prev`, but they could be in the same bfs layer so it's not always the case.

app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
ggawryal: I see, maybe you can rename `update` to `updateResultOrSetOnWhichResultDepends` or something similar to emphasize the reason behind setting `wasResultUpdated` flag in this function

PR #28, opened 2022-10-25
Add the outline of the new parser classes

rkilar:


app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -0,0 +1,10 @@
+package compiler
+
+import compiler.common.diagnostics.Diagnostics
+import java.io.Reader
+
+class Compiler(val diagnostics: Diagnostics) {
```
tomekb234, line 6: I think we should use an interface for diagnostics here.
We can either replace `Diagnostics` with `(Diagnostic) -> Unit` here, or turn `Diagnostics` into a functional interface.
https://kotlinlang.org/docs/fun-interfaces.html

app/src/main/kotlin/compiler/parser/Parser.kt
```diff
@@ -0,0 +1,32 @@
...
+import compiler.common.diagnostics.Diagnostics
+import compiler.parser.grammar.AutomatonGrammar
+import compiler.parser.grammar.Grammar
+import compiler.parser.grammar.Production
+
+class Parser<S : Comparable<S>>(
+    val automatonGrammar: AutomatonGrammar<S>,
+    val nullable: Set<S>,
+    val first: Map<S, Set<S>>,
+    val follow: Map<S, Set<S>>,
```
tomekb234, line 13: I'd suggest packing these three into a single data class, e.g. `GrammarAnalysis.Result`.

app/src/main/kotlin/compiler/parser/Parser.kt
```diff
@@ -0,0 +1,32 @@
...
+    val diagnostics: Diagnostics
+) {
+    val parseActions: Map<Pair<DfaState<S, Production<S>>, S>, Action<S>>
```
tomekb234, line 16: The key type should be `Triple<Dfa<S, Production<S>>, DfaState<S, Production<S>>, S>`, unless the states of distinct DFAs are guaranteed to be different. This is currently not the case for `RegexDfaState`, which is a `data class`, and so it is possible for two DFAs to coincidentally have the same state (in the sense of the `equals` method).

app/src/main/kotlin/compiler/parser/Parser.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    init {
+        // TODO: Compute the parsing table
+        parseActions = emptyMap()
+    }
+
+    companion object {
+        fun <S : Comparable<S>> create(grammar: Grammar<S>): Parser<S> {
```
tomekb234, line 24: Maybe just a secondary constructor instead? Also, we need to pass `Diagnostics` here.

app/src/main/kotlin/compiler/parser/Action.kt
```diff
@@ -0,0 +1,11 @@
+package compiler.parser
+
+import compiler.parser.grammar.Production
+
+sealed class Action<S : Comparable<S>> {
```
tomekb234, line 5: Maybe less generally sounding `ParserAction`?

tomekb234 (approved):

PR #35, opened 2022-10-29
Implement the parser's process method

tomekb234:
~The tests are yet to be written, after the parsing table generation is implemented.~

### DIFF
--- a/app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
+++ b/app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
@@ -1,5 +1,7 @@
 package compiler.common.diagnostics
 
+import compiler.lexer.Location
+
 sealed class Diagnostic {
     abstract fun isError(): Boolean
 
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -8,4 +10,22 @@ sealed class Diagnostic {
             TODO("Not yet implemented")
         }
     }
+
+    class ParserError(
+        val symbol: String?,
+        val start: Location,
+        val end: Location,
+        val expectedSymbols: List<String>
```
rkilar, line 18: Why store symbols as Strings instead of actual symbols?

rkilar (changes requested):
The code loooks good. I only have a single question regarding ParserError

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
tomekb234: Because the symbol's type is abstract and in the end we only care about its string representation. On the other hand, sure, it is technically more efficient to _store_ the object itself and only convert it to `String` when needed.

tomekb234: Changed to `Any` instead of `String`. The class could be generic on the symbol's type, but we wouldn't gain anything from this.

rkilar (approved):

PR #37, opened 2022-11-01
Add initial end-to-end tests

ignacy123:
There are some workarounds for the not implemented parser. This will be merged after the parser implementation.
The amount of tests that fail during lexical analysis is insignificant, as it is very hard to come up with reasonable examples. Most of the "wrong" constructs will be just read as `TokenType.Identifier`, as the regex for this token accepts "almost" anything.

### DIFF
--- a/app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
+++ b/app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
@@ -1,0 +1,182 @@
+package compiler.e2e
+
+import kotlin.test.Test
+
+class CorrectPrograms {
+    @Test
+    fun `test identifiers`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            zm n: Liczba = 10234
+            zm camelCaseIdentifier: Liczba = 10234
+            zm snake_case_identifier: Liczba = 10234
+            zm ident1f13r_w1tH_n00mb3r5: Liczba = 10234
+            """
+        )
+    }
+
+    @Test
+    fun `test types and literals`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            zm n: Liczba = 10234
+            n = 0
+            n = -0
+            n = -56789
+            n = +56789
+            n = 2147483647 // 32 bitowa ze znakiem
+            n = -2147483648
+
+            zm b: Czy = prawda
+            b = fałsz
+            """
+        )
+    }
+
+    @Test
+    fun `test if, elif and else`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            zm x: Liczba
+            zm y: Liczba = 10
+            jeśli (fałsz) {
+                x = 1
+            }
+            zaś gdy (y == 10) {
+                x = 2
+            }
+            wpp {
+                x = 3
+            }
+            napisz(x) //wypisze 2
+            jeśli(x == 2) {
+                napisz(14)
+            }
+            """
+        )
+    }
+
+    @Test
+    fun `test while, break and continue`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            zm x: Liczba = 0
+            dopóki (prawda) { //dopóty
+                jeśli (x == 420) pomiń
+                napisz(x)
+                x=x+1
+                jeśli (x == 1000) przerwij
+            }
+            """
+        )
+    }
+
+    @Test
+    fun `test return`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            czynność przykład() -> Liczba {
+                zwróć 42
+            }
+
+            czynność przykład2() {
+                zm x: Liczba = 0
+                dopóki (prawda) {
+                    napisz(x)
+                    x=x+1
+                    jeśli(x==10) zakończ
+                }
+            }
+            """
+        )
+    }
+
+    @Test
+    fun `test default arguments`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            czynność suma(a: Liczba, b: Liczba = 2, c: Liczba = 4, d: Liczba = 8) -> Liczba {
+                zwróć a + b + c + d
+            }
+
+            suma(10, 10, 10, 10) // 40
+            suma(10, 10, 10)     // 38 = 10 + 10 + 10 + 8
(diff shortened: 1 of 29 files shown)


app/src/test/kotlin/compiler/e2e/common/E2eAsserter.kt
```diff
@@ -0,0 +1,18 @@
...
+
+object E2eAsserter {
+    fun assertProgramGeneratesDiagnostics(program: String, diagnostics: List<Diagnostic>) {
+        val actualDiagnostics = CompilerDiagnostics()
+        val compiler = Compiler(actualDiagnostics)
+        compiler.process(program.trimIndent().reader())
+        assertContentEquals(diagnostics.asSequence(), actualDiagnostics.diagnostics)
+    }
+    fun assertProgramCorrect(program: String) {
+        assertProgramGeneratesDiagnostics(program, listOf())
```
rkilar, line 16: Technically, a program could be correct and still generate non-error diagnostics. Maybe we could change the name or only check for errors here. It's probably not that important as we're probably not gonna generate many warnings, certainly not in current tests.

rkilar (changes requested):
Not much to check here. The only problem is that you forgot to filter out whitespace tokens from Lexer output.

app/src/test/kotlin/compiler/e2e/common/E2eAsserter.kt
ignacy123: I think that there aren't enough tests for now to pick a good naming convention :thinking:. I assume we'll be adding more of those tests this week, so let's postpone the choice of the naming convention until then.

app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
```diff
@@ -0,0 +1,182 @@
...
+
+class CorrectPrograms {
+    @Test
+    fun `test identifiers`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            zm n: Liczba = 10234
+            zm camelCaseIdentifier: Liczba = 10234
+            zm snake_case_identifier: Liczba = 10234
+            zm ident1f13r_w1tH_n00mb3r5: Liczba = 10234
```
rkilar, line 13: Newline at the end of file is needed to parse correctly

app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    @Test
+    fun `test types and literals`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            zm n: Liczba = 10234
+            n = 0
+            n = -0
+            n = -56789
+            n = +56789
```
rkilar, line 26: This breaks as there is no unary plus operator in the grammar. You can add it to the grammar or remove this line.

app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
```diff
@@ -0,0 +1,182 @@
...
+            n = 2147483647 // 32 bitowa ze znakiem
+            n = -2147483648
+
+            zm b: Czy = prawda
+            b = fałsz
```
rkilar, line 31: Global scope can only contain function or global variable declarations. Wrap this in a function.

app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
```diff
@@ -0,0 +1,182 @@
...
+    fun `test while, break and continue`() {
+        E2eAsserter.assertProgramCorrect(
+            """
+            zm x: Liczba = 0
+            dopóki (prawda) { //dopóty
+                jeśli (x == 420) pomiń
+                napisz(x)
+                x=x+1
+                jeśli (x == 1000) przerwij
+            }
```
rkilar, line 69: Here as well.

app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
```diff
@@ -0,0 +1,182 @@
...
+            """
+            czynność suma(a: Liczba, b: Liczba = 2, c: Liczba = 4, d: Liczba = 8) -> Liczba {
+                zwróć a + b + c + d
+            }
+
+            suma(10, 10, 10, 10) // 40
+            suma(10, 10, 10)     // 38 = 10 + 10 + 10 + 8
+            suma(0)              // 14 = 0 + 2 + 4 + 8
+            suma(10, 11, d=0)    // 10 + 11 + 4 + 0
+            suma(0, c=100)       // 110 = 0 + 2 + 100 + 8
```
rkilar, line 106: Same for these function calls.

app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
```diff
@@ -0,0 +1,182 @@
...
+                    zwróć 1
+                wpp
+                    zwróć n * silnia(n-1)
+            }
+            """
+        )
+    }
+
+    @Test
+    fun `test variables, values and constants`() {
```
rkilar, line 146: Again needs function.

app/src/test/kotlin/compiler/e2e/LexicallyIncorrectPrograms.kt
```diff
@@ -0,0 +1,78 @@
+package compiler.e2e
+
+import compiler.common.diagnostics.Diagnostic
+import compiler.lexer.Location
+import kotlin.test.Test
+
+class LexicallyIncorrectPrograms {
```
rkilar, line 7: These test fail because the incorrect code from Lexer causes further errors in Parser.

rkilar (changes requested):
Some of the code examples need to be fixed to comply with the grammar. Parser errors need to be ignored in lexer tests. Once parser table is merged these tests should pass.

app/src/test/kotlin/compiler/e2e/CorrectPrograms.kt
ignacy123: I just removed the plus, we don't need it

rkilar (approved):

PR #39, opened 2022-11-03
Parser table

woocash2:
This is a parser table pull request


app/src/main/kotlin/compiler/parser/analysis/GrammarAnalysis.kt
```diff
@@ -184,4 +165,20 @@ class GrammarAnalysis<S : Comparable<S>> {
...
+            firstPlus.putIfAbsent(symbol, HashSet())
+            firstPlus[symbol]!!.addAll(firstOfSymbol)
+        }
+
+        for (symbol in nullable) {
+            firstPlus.putIfAbsent(symbol, HashSet())
+            firstPlus[symbol]!!.addAll(follow[symbol]!!)
+        }
+
+        return firstPlus
```
rkilar, line 182: I feel like this function could be shorter and more readable with something like `return grammar.getSymbols().associateWith` and `union` to closer match the mathematical definition.

rkilar (changes requested):
The code looks to be correct. I left some small stylistic suggestions.

bebidek:
I love the commit message "Shut up, ktlint".

rkilar (approved):

PR #40, opened 2022-11-05
Implement language grammar

bebidek:


app/src/main/kotlin/compiler/parser/grammar/ParserGrammar.kt
```diff
@@ -1,7 +1,109 @@
...
+            getProduction("EXPR8", "\\{nEXPR16}$NLS\\{tIFF}$NLS\\{nEXPR8}"),
+
+            getProduction("EXPR4", "\\{nEXPR8}"),
+            getProduction("EXPR4", "\\{nEXPR8}$NLS\\{tAND}$NLS\\{nEXPR4}"),
+
+            getProduction("EXPR2", "\\{nEXPR4}"),
+            getProduction("EXPR2", "\\{nEXPR4}$NLS\\{tOR}$NLS\\{nEXPR2}"),
+
+            getProduction("EXPR", "\\{nEXPR2}"),
+            getProduction("EXPR", "\\{nEXPR2}$NLS\\{tQUESTION_MARK}$NLS\\{nEXPR2}$NLS\\{tCOLON}$NLS\\{nEXPR}"),
```
rkilar, line 83: I've been trying to run this through the Parser from existing PRs and I keep getting errors. Let's say we want to deduce `EXPR` from sequence `EXPR2, NEWLINE, ...`. After reading `EXPR2` we can either shift (consume `NEWLINE` and step in the dfa) or reduce with production `EXPR -> EXPR2` as `NEWLINE` is part of `folllow[EXPR]`. Am I understanding this wrong so is the Parser or is this grammar not LL(1)?

rkilar (changes requested):

app/src/main/kotlin/compiler/parser/grammar/ParserGrammar.kt
bebidek: You're right, the grammar is not LL(1). I managed to solve the problem with EXPR and NEWLINE you have found (although it's an ugly solution), but another one came out. We'll try to fix it on Sunday.

rkilar: Yeah, it was just an example. The parser returned many similar errors, mostly based on the newline symbol.

bebidek:
Now the grammar seems to be LL(1) (at least parser table from #39 is generated without exceptions). However, it became ugly (lots of copy-paste fragments I can't get rid of).

app/src/main/kotlin/compiler/parser/grammar/Symbol.kt
```diff
@@ -5,13 +5,13 @@ import compiler.lexer.lexer_grammar.TokenType
...
-            TODO("Not yet implemented")
+            return if (other is Terminal) tokenType.compareTo(other.tokenType) else -1
         }
     }
 
     data class NonTerminal(val nonTerminal: NonTerminalType) : Symbol() {
         override fun compareTo(other: Symbol): Int {
-            TODO("Not yet implemented")
+            return if (other is NonTerminal) nonTerminal.compareTo(other.nonTerminal) else 1
         }
```
rkilar, line 15: I feel like it could be useful to have a more concise toString function that just returns the string for tokenType/nonTerminal, maybe with n/t as opposed to the default from data class.

app/src/test/kotlin/compiler/parser/grammar/ParserGrammarTest.kt
```diff
@@ -0,0 +1,9 @@
+package compiler.parser.grammar
+
+import kotlin.test.Test
+
+class ParserGrammarTest {
+    @Test fun `test grammar regexes are correct syntactically`() {
+        ParserGrammar.getGrammar()
+    }
```
rkilar, line 8: Maybe a test where we run the grammar through the parser contructor could be useful so we know it doesn't break. Could be in this file or in the parser.integration_tests package.

rkilar (changes requested):
Looks good as far as I can tell. I only have a question about one production and couple simple suggestions.

app/src/main/kotlin/compiler/parser/grammar/ParserGrammar.kt
```diff
@@ -1,7 +1,170 @@
...
+            getProduction("CONST", "{tINTEGER}|{tTRUE_CONSTANT}|{tFALSE_CONSTANT}"),
+
+            getProduction("VAR_DECL", "({tVARIABLE}|{tVALUE}|{tCONSTANT})$NLS{tIDENTIFIER}$NLS{tCOLON}$NLS{nTYPE}({tASSIGNMENT}$NLS{nEXPR})?"),
+            getProduction("FUNC_DEF", "{tFUNCTION}$NLS{tIDENTIFIER}$NLS{tLEFT_PAREN}$NLS{nDEF_ARGS}{tRIGHT_PAREN}($NLS{tARROW}$NLS{nTYPE})?{tLEFT_BRACE}{nMANY_STATEMENTS}{tRIGHT_BRACE}"),
+
+            getProduction("DEF_ARGS", "({nDEF_ARG}($NLS{tCOMMA}$NLS{nDEF_ARG})*$NLS({tCOMMA}$NLS{nDEF_ARG}$NLS{tASSIGNMENT}$NLS{nE_EXPR})*)?"),
+            getProduction("DEF_ARGS", "{nDEF_ARG}$NLS{tASSIGNMENT}$NLS{nE_EXPR}({tCOMMA}$NLS{nDEF_ARG}$NLS{tASSIGNMENT}$NLS{nE_EXPR})*"),
+            getProduction("DEF_ARG", "{tIDENTIFIER}$NLS{tCOLON}$NLS{nTYPE}"),
+            getProduction("CALL_ARGS", "({nE_EXPR}({tCOMMA}$NLS{nE_EXPR})*({tCOMMA}$NLS{nE_EXPR}{tASSIGNMENT}$NLS{nE_EXPR})*)?"),
+            getProduction("CALL_ARGS", "{nE_EXPR}{tASSIGNMENT}$NLS{nE_EXPR}({tCOMMA}$NLS{nE_EXPR}{tASSIGNMENT}$NLS{nE_EXPR})*"),
```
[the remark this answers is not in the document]
bebidek: Yes, it should, but LL(1) can't handle this (at least in this form), because when you see "f(5, " and the next symbol is identifier "x" you don't know whether it is "f(5, x + 3)" or "f(5, x = 8)". The same story with:
`getProduction("ATOMIC_STATEMENT", "{nEXPR}{tASSIGNMENT}$NLS{nEXPR}"),`
This will have to be resolved at next phase (luckily, it's going to be an easy check).

app/src/main/kotlin/compiler/parser/grammar/Symbol.kt
bebidek: I didn't suffer much from the default toString from data class during debugging, but custom one can be added too.

app/src/main/kotlin/compiler/parser/grammar/ParserGrammar.kt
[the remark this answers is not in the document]
rkilar: I see, makes sense.

rkilar (approved):

PR #42, opened 2022-11-08
Create semantic analysis templates

kamgal00:


app/src/test/kotlin/compiler/parser/integration_tests/ParserErrorsTest.kt
```diff
@@ -0,0 +1,5 @@
+package compiler.parser.integration_tests
+
+class ParserErrorsTest {
```
JHodorHodor, line 3: Isn't this supposed to be in `e2e` directory? (And same for all the below)

app/src/main/kotlin/compiler/ast/AstNodeTypes.kt
```diff
@@ -0,0 +1,6 @@
+package compiler.ast
+
+object AstNodeTypes {
```
tomekb234, line 3: Is this `object` necessary? The interfaces below can be top-level (and we can assume a convention that their names end with `Node`).

app/src/main/kotlin/compiler/semantic_analysis/NameResolver.kt
```diff
@@ -0,0 +1,12 @@
+package compiler.semantic_analysis
+
+import compiler.ast.AstNodeTypes
+import compiler.ast.Program
+import compiler.common.diagnostics.Diagnostics
+import compiler.common.semantic_analysis.ReferenceHashMap
+
+object NameResolver {
+    fun calculateNameResolution(ast: Program, diagnostics: Diagnostics): ReferenceHashMap<Any, AstNodeTypes.NamedNode> {
```
tomekb234, line 9: This can return just `ReferenceMap` (the same is with the other `calculate` functions).

app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
```diff
@@ -0,0 +1,15 @@
...
+import compiler.ast.AstNodeTypes
+import compiler.ast.Program
+import compiler.common.diagnostics.Diagnostics
+import compiler.common.semantic_analysis.ReferenceHashMap
+import compiler.common.semantic_analysis.ReferenceMap
+
+object VariablePropertiesAnalyzer {
+    data class VariableProperties(val owner: AstNodeTypes.FunctionDefinition, val usedInNested: Boolean)
+
+    fun calculateVariablesProperties(ast: Program, nameResolution: ReferenceMap<Any, AstNodeTypes.NamedNode>, diagnostics: Diagnostics): ReferenceHashMap<Any, VariableProperties> {
```
tomekb234, line 12: `calculateVariableProperties` sounds better.

app/src/main/kotlin/compiler/ast/Program.kt
```diff
@@ -2,7 +2,7 @@ package compiler.ast
 
 data class Program(val globals: List<Global>) {
     sealed class Global {
-        data class VariableDefinition(val variable: Variable) : Global()
-        data class FunctionDefinition(val function: Function) : Global()
+        data class VariableDefinition(val variable: Variable) : Global(), AstNodeTypes.NamedNode
```
tomekb234, line 5: Is it necessary for `NamedNode`s to point to `Variable` wrappers such as `Program.VariableDefinition` and `Statement.VariableDefinition`, and not just to `Variable`s itself? Same question for the `FunctionDefinition` interface - can't `VariableProperties` objects just point directly to `Function`s?

tomekb234 (approved):

PR #43, opened 2022-11-10
Name resolution errors

woocash2:


app/src/main/kotlin/compiler/semantic_analysis/NameResolver.kt
```diff
@@ -4,8 +4,16 @@ import compiler.ast.NamedNode
 import compiler.ast.Program
 import compiler.common.diagnostics.Diagnostics
 import compiler.common.semantic_analysis.ReferenceMap
+import java.lang.Exception
 
 object NameResolver {
+
+    sealed class NameResolutionError {
```
JHodorHodor, line 11: This should be added to `Diagnostic.kt` file.

PR #45, opened 2022-11-10
Name resolution errors e2e tests

JHodorHodor:
All new tests have `@Ignore` annotation! I started with a branch of the parser tests, so probably it is the best to compare changes with it.

### DIFF
--- a/app/src/test/kotlin/compiler/e2e/NameResolutionErrorsTest.kt
+++ b/app/src/test/kotlin/compiler/e2e/NameResolutionErrorsTest.kt
@@ -1,5 +1,246 @@
 package compiler.e2e
 
+import compiler.common.diagnostics.Diagnostic
+import compiler.e2e.E2eAsserter.assertErrorOfType
+import kotlin.test.Ignore
+import kotlin.test.Test
+
 class NameResolutionErrorsTest {
-    // TODO: implement tests for name resolution errors
+
+    @Ignore
+    @Test
+    fun `test undefined variable`() {
+        assertErrorOfType(
+            """
+                    czynność f() {
+                        x = 17
+                    }
+                    
+                """,
+            Diagnostic.NameResolutionErrors.UndefinedVariable::class
+        )
+    }
+
+    @Ignore
+    @Test
+    fun `test variable defined in other scope`() {
+        assertErrorOfType(
+            """
+                    czynność f() -> Liczba {
+                        {
+                            zm x: Liczba = 17
+                        }
+                        zwróć x
+                    }
+                    
+                """,
+            Diagnostic.NameResolutionErrors.UndefinedVariable::class
+        )
+    }
+
+    @Ignore
+    @Test
+    fun `test variable defined in an inner function`() {
+        assertErrorOfType(
+            """
+                    czynność f() -> Liczba {
+                        czynność g() {
+                            zm x: Liczba = 17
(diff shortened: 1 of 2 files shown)


ignacy123:
I think it is also worth adding the following test case:
```
x = 15
zm x: Liczba = 17
```
to ensure we don't mess up the order somewhere. Maybe the same idea can be used for other scenarios?

app/src/test/kotlin/compiler/e2e/NameResolutionErrorsTest.kt
```diff
@@ -1,5 +1,246 @@
...
+
+    // I'm not 100% sure if the next two tests make sense
+    @Ignore
+    @Test
+    fun `test using functions in assignment`() {
+        assertErrorOfType(
+            """
+                    czynność f() {
+                        czynność g() { }
+                        zm x: Liczba = x
```
kamgal00, line 224: I think it should be `zm x: Liczba = g`

app/src/test/kotlin/compiler/e2e/NameResolutionErrorsTest.kt
```diff
@@ -1,5 +1,246 @@
...
+                        zm x: Czy = prawda
+                        zwróć x()
+                    }
+                    
+                """,
+            Diagnostic.NameResolutionErrors.VariableIsNotCallable::class
+        )
+    }
+
+    // I'm not 100% sure if the next two tests make sense
```
kamgal00, line 216: In my opinion, they make sense - our language doesn't support using functions as variables at this point.

app/src/test/kotlin/compiler/e2e/NameResolutionErrorsTest.kt
```diff
@@ -1,5 +1,246 @@
...
+            """
+                    czynność f() -> Czy {
+                        czynność g() { }
+                        zwróć (g == 17)
+                    }
+                    
+                """,
+            Diagnostic.NameResolutionErrors.FunctionIsNotVariable::class
+        )
+    }
```
kamgal00, line 245: Could you please add some tests for function parameters? We could test, for example, name conflicts (between two parameters and between a parameter and a variable), treating parameters as callables, using parameters outside function scope, and maybe using undefined variables/functions as default parameter values:
```
czynność f(x: Liczba = y) {}
czynność f(x: Liczba = f()) {}
```

app/src/test/kotlin/compiler/e2e/NameResolutionErrorsTest.kt
```diff
@@ -1,5 +1,246 @@
...
+    fun `test using functions in conditions`() {
+        assertErrorOfType(
+            """
+                    czynność f() -> Czy {
+                        czynność g() { }
+                        zwróć (g == 17)
+                    }
+                    
+                """,
+            Diagnostic.NameResolutionErrors.FunctionIsNotVariable::class
```
kamgal00, line 243: This test is probably redundant - the previous one already tests if `g` is a correct expression

kamgal00 (changes requested):
Testy są dobrze napisane i rozpatrują wszystkie przypadki, brakuje mi jedynie kilku testów dla parametrów funkcji.

kamgal00 (approved):

PR #53, opened 2022-11-13
Implement the type checker

tomekb234:


app/src/main/kotlin/compiler/semantic_analysis/TypeChecker.kt
```diff
@@ -1,14 +1,247 @@
...
+                        Expression.BinaryOperation.Kind.EQUALS,
+                        Expression.BinaryOperation.Kind.NOT_EQUALS,
+                        Expression.BinaryOperation.Kind.LESS_THAN,
+                        Expression.BinaryOperation.Kind.LESS_THAN_OR_EQUALS,
+                        Expression.BinaryOperation.Kind.GREATER_THAN,
+                        Expression.BinaryOperation.Kind.GREATER_THAN_OR_EQUALS -> {
+                            checkExpression(expression.leftOperand, Type.Number)
+                            checkExpression(expression.rightOperand, Type.Number)
+                            return Type.Number
+                        }
```
bebidek, line 188: Shouldn't these give Type.Boolean?

tomekb234: Yeah, they should.

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -32,4 +37,59 @@ sealed class Diagnostic {
...
+
+    data class VariableCall(val call: Expression.FunctionCall, val variable: Variable) : Diagnostic() {
+        override fun isError() = true
+        override fun toString() = "Cannot call a variable"
+    }
+
+    data class ParameterCall(val call: Expression.FunctionCall, val parameter: Function.Parameter) : Diagnostic() {
+        override fun isError() = true
+        override fun toString() = "Cannot call a parameter"
+    }
```
kamgal00, line 79: Do we need those diagnostics here? I think these errors are already checked in name resolution

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -32,4 +37,59 @@ sealed class Diagnostic {
...
+
+    data class NonConstantExpression(val expression: Expression) : Diagnostic() {
+        override fun isError() = true
+        override fun toString() = "Expected a constant expression"
+    }
+
+    data class InvalidType(val expression: Expression, val type: Type, val expectedType: Type) : Diagnostic() {
+        override fun isError() = true
+        override fun toString() = "Expected type '$expectedType' instead of '$type'"
+    }
```
kamgal00, line 94: To remain consistent with the others, could you put those diagnostics in some sealed class, e. g. `TypeCheckingError`?

kamgal00 (approved):
Wszystko wygląda w porządku, dodałem dwa małe pytania

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
tomekb234: I think it should actually be the responsibility of the type checker, not the name resolution, to report these errors, because they are errors only for as long as functions are not first-class, and this is a property of the type system. In principle, variables, parameters and functions have the same namespace, and name resolution shouldn't have to care in what contexts these names are used.

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
tomekb234: Done.

tomekb234: Done.

PR #54, opened 2022-11-13
Name resolution

woocash2:


app/src/main/kotlin/compiler/semantic_analysis/NameResolver.kt
```diff
@@ -1,12 +1,269 @@
...
+                    analyzeNode(node.variable, currentScope)
+                    addName(node.variable.name, node.variable, currentScope)
+                }
+
+                is Statement.FunctionDefinition -> {
+                    reportIfNameConflict(node.function.name, currentScope)
+
+                    // first add name, then analyze because we can have recursive calls in the body
+                    addName(node.function.name, node.function, currentScope)
+                    analyzeNode(node.function, currentScope)
```
kamgal00, line 224: Cases `Statement.FunctionDefinition` and `Global.FunctionDefinition` are basically the same - could we move the repeated code to the `Function` case and here simply write `is Statement.FunctionDefinition -> analyzeNode(node.function, currentScope)`? Could we do the same for `Statement.VariableDefinition`/`Global.VariableDefinition`?

kamgal00 (approved):
Kod jest czytelny, zgrabnie napisany i dobrze opisany komentarzami :) dodałem jedną małą uwagę, tak to chyba wszystko git

app/src/main/kotlin/compiler/semantic_analysis/NameResolver.kt
```diff
@@ -1,12 +1,269 @@
...
+                }
+
+                is Statement.Assignment -> {
+                    reportIfVariableUndefined(node.variableName)
+                    reportIfFunctionUsedAsVariable(node.variableName) // when we try to assign to a function
+                    nameDefinitions[node.variableName] = visibleNames[node.variableName]!!.topVariable()
```
rkilar, line 230: Shouldn't the key be be `node` not `node.variableName`?

app/src/main/kotlin/compiler/semantic_analysis/NameResolver.kt
```diff
@@ -1,12 +1,269 @@
...
+                    analyzeNode(node.variable, currentScope)
+                    addName(node.variable.name, node.variable, currentScope)
+                }
+
+                is Program.Global.FunctionDefinition -> {
+                    reportIfNameConflict(node.function.name, currentScope)
+
+                    // first add name, then analyze because we can have recursive calls in the body
+                    addName(node.function.name, node.function, currentScope)
+                    analyzeNode(node.function, currentScope)
```
kamgal00, line 146: There is a small problem here: we allow self-reference in default parameter values: `czynność f(x: Czy = f())`

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -32,4 +32,14 @@ sealed class Diagnostic {
...
     }
+
+    sealed class NameResolutionErrors() : Diagnostic() {
+        override fun isError() = true
+
+        class UndefinedVariable : NameResolutionErrors()
+        class UndefinedFunction : NameResolutionErrors()
+        class NameConflict : NameResolutionErrors()
+        class VariableIsNotCallable : NameResolutionErrors()
+        class FunctionIsNotVariable : NameResolutionErrors()
```
tomekb234, line 43: https://github.com/TCS-Compilers/compiler/pull/53#discussion_r1020980440

tomekb234 (commented):
`gradle build` runs ktlint and shows its errors, so you can use this before pushing and not have GitHub reject your commits.

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -32,4 +32,14 @@ sealed class Diagnostic {
                 append(" Expected symbols: ${expectedSymbols.joinToString()}.")
         }.toString()
     }
+
+    sealed class NameResolutionErrors() : Diagnostic() {
+        override fun isError() = true
+
+        class UndefinedVariable : NameResolutionErrors()
```
JHodorHodor, line 39: Don't you think that we should print some more info, like in parser/lexer? I.e. name of the variable, line, etc.?

app/src/main/kotlin/compiler/semantic_analysis/NameResolver.kt
woocash2: good point

app/src/main/kotlin/compiler/semantic_analysis/NameResolver.kt
woocash2: Right, I think I can move adding names from `FunctionDefinition` to `Function` and from `VariableDefinition` to `Variable`

PR #55, opened 2022-11-13
Add end-to-end tests for type checking and variable property analysis

rkilar:


app/src/test/kotlin/compiler/e2e/TypeCheckingErrorsTest.kt
```diff
@@ -1,5 +1,760 @@
...
+        assertErrorOfType(program, Diagnostic.ConditionalTypesMismatch::class)
+    }
+
+    fun assertUninitializedGlobalVariableError(program: String) {
+        assertErrorOfType(program, Diagnostic.UninitializedGlobalVariable::class)
+    }
+
+    fun assertConstantWithoutValueError(program: String) {
+        assertErrorOfType(program, Diagnostic.ConstantWithoutValue::class)
+    }
```
kamgal00, line 23: Those functions can be private, as well as assert functions in `VariablePropertiesAnalysisError`

app/src/test/kotlin/compiler/e2e/TypeCheckingErrorsTest.kt
```diff
@@ -1,5 +1,760 @@
...
+
+    @Ignore
+    @Test
+    fun `test define variable with unit type`() {
+        assertInvalidTypeError("zm a: Nic;")
+        assertInvalidTypeError("wart a: Nic;")
+        assertInvalidTypeError("stała a: Nic;")
```
kamgal00, line 30: This test may also throw diagnostic `UninitializedGlobalVariable`. Maybe put those variables inside some functions?

kamgal00 (changes requested):
Testy są wyczerpujące, wszystko wygląda ok :) wrzucam dwie małe uwagi

app/src/test/kotlin/compiler/e2e/TypeCheckingErrorsTest.kt
```diff
@@ -1,5 +1,927 @@
...
+    private fun assertNonConstantExpressionError(program: String) {
+        assertErrorOfType(program, Diagnostic.NonConstantExpression::class)
+    }
+
+    @Ignore
+    @Test
+    fun `test define variable with unit type`() {
+        assertInvalidTypeError("czynność test() { zm a: Nic; }")
+        assertInvalidTypeError("czynność test() { wart a: Nic; }")
+        assertInvalidTypeError("czynność test() { stała a: Nic; }")
```
kamgal00, line 35: Sorry I didn't write it before, but I think variables and properties with type `Nic` are legal after all: they have only one possible value `nic`. Could we remove this test, as well as the next one?

rkilar:
I added tests for `NonConstantExpression` errors. Right now we only consider integer and boolean literals to be constant so some cases are commented out. If we ever add more accurate checks, they can be added back.

kamgal00 (approved):
Wszystko jest ok, przy mergowaniu pamiętaj tylko żeby na razie zostawić Ignore przy `test wrong function call argument type`

PR #56, opened 2022-11-15
Templates for intermediate form generation.

woocash2:


app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -0,0 +1,16 @@
+package compiler.semantic_analysis
+
+import compiler.ast.Expression
+import compiler.ast.Function
+import compiler.ast.Program
+import compiler.common.diagnostics.Diagnostics
+import compiler.common.semantic_analysis.ReferenceMap
+
+object ArgumentResolver {
+    fun calculateArgumentToParameterResolution(
```
JHodorHodor, line 10: Can you also add this to Compiler.kt?

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -38,6 +38,10 @@ sealed class Diagnostic {
         }.toString()
     }
 
+    sealed class ArgumentResolutionErrors() {
```
JHodorHodor, line 41: This should extend Diagnostic. Also should't this be after NameResolutionErrors?

app/src/main/kotlin/compiler/intermediate_form/FunctionDependenciesAnalyzer.kt
```diff
@@ -0,0 +1,16 @@
+package compiler.intermediate_form
+
+import compiler.ast.Function
+import compiler.ast.Program
+import compiler.ast.Variable
+import compiler.common.semantic_analysis.ReferenceMap
+
+object FunctionDependenciesAnalyzer {
+    fun variablesUsedByFunctions(ast: Program): ReferenceMap<Function, Pair<List<Variable>, List<Variable>>> { // (accessed, modified)
```
JHodorHodor, line 9: I think we shouldn't use Pair here, but some simple wrapper data class with named fields instead

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraph.kt
```diff
@@ -0,0 +1,15 @@
+package compiler.intermediate_form
+
+import compiler.common.semantic_analysis.ReferenceMap
+
+typealias Node = IntermediateFormTreeNode
```
JHodorHodor, line 5: Maybe `IFTNode`?

JHodorHodor (approved):

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
tomekb234: Also, the name should be singular `ArgumentResolutionError`.

app/src/main/kotlin/compiler/intermediate_form/FunctionDependenciesAnalyzer.kt
tomekb234: I would suggest something like `ReferenceMap<Function, ReferenceMap<Variable, VariableAccessMode>>` with `VariableAccessMode` being a two-option enum.

app/src/main/kotlin/compiler/intermediate_form/FunctionDependenciesAnalyzer.kt
```diff
@@ -0,0 +1,16 @@
...
+        return TODO()
+    }
+
+    fun createCallGraph(ast: Program): Set<Pair<Function, Function>> {
```
tomekb234, line 13: A problem here is that we need to compare `Function`s by reference. This should be something like `ReferenceMap<Function, ReferenceSet<Function>>`.

tomekb234: A `ReferenceSet` can be made in this way:
```kotlin
interface ReferenceSet<E> : Set<E>

interface MutableReferenceSet<E> : MutableSet<E>, ReferenceSet<E>

class ReferenceHashSet<E> : MutableSet<E> by Collections.newSetFromMap(ReferenceHashMap()), MutableReferenceSet<E>
```

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -0,0 +1,16 @@
...
+        ast: Program,
+        diagnostics: Diagnostics
+    ): ReferenceMap<Expression.FunctionCall.Argument, Function.Parameter> {
```
tomekb234, line 13: A type alias `ArgumentResolutionResult` may be convenient.

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
woocash2: I matched it to `NameResolutionErrors`. I think I'll convert this to singular as well.

app/src/main/kotlin/compiler/intermediate_form/FunctionDependenciesAnalyzer.kt
woocash2: Ok, then we can't have info like "modifies but doesn't read", but I think it's not necessary. ReadOnly and ReadWrite should be enough for our future usecases.

PR #63, opened 2022-11-18
Update the calculation of variable properties

ignacy123:
We haven't agreed on the particular style of mapping (`Function` -> `Parameter` or `Parameter` -> `Function`), but we can easily transition between the two if there will ever be a need to do so.

### DIFF
--- a/app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
+++ b/app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
@@ -26,74 +26,63 @@ object VariablePropertiesAnalyzer {
         diagnostics: Diagnostics
     ): ReferenceMap<Any, VariableProperties> {
         val variableProperties: MutableReferenceMap<Any, VariableProperties> = ReferenceHashMap()
-        fun resolve(
-            // Any = Statement.Assignment | Expression.Variable
-            node: Any,
-            currentOwner: Function?,
-            isAssignment: Boolean
-        ) {
-            val resolvedVariable: Any = nameResolution.get(node)!!
-            // we assume variable is defined before being referenced
-            // if this fails then something is wrong with name resolution
-            val oldOwner = variableProperties.get(resolvedVariable)!!.owner
-            if (resolvedVariable is Function.Parameter && isAssignment) {
-                // function parameter must have an owner
-                diagnostics.report(AssignmentToFunctionParameter(resolvedVariable, oldOwner!!, currentOwner!!))
-            }
-            if (oldOwner != currentOwner) {
-                // if this was a function that is not nested in the original owner
-                // then name resolution would not be able to succeed
-                // current owner cannot be null as it is always a function once we start descending
-                if (isAssignment)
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
```diff
@@ -26,74 +26,63 @@ object VariablePropertiesAnalyzer {
...
+                }
+                is Statement.FunctionDefinition -> analyzeVariables(node.function, currentFunction)
+                is Global.FunctionDefinition -> analyzeVariables(node.function, currentFunction)
                 is Statement.Assignment -> {
-                    resolve(node, currentOwner, true)
-                    analyzeVariables(node.value, currentOwner)
+                    val resolvedVariable: NamedNode = nameResolution[node]!!
+                    variableProperties[resolvedVariable]!!.writtenIn.add(currentFunction!!)
+                    if (resolvedVariable is Function.Parameter) {
+                        diagnostics.report(
```
woocash2, line 48: Do we disallow modifying outer function's parameters?

app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
```diff
@@ -26,74 +26,63 @@ object VariablePropertiesAnalyzer {
...
+                            AssignmentToFunctionParameter(
+                                resolvedVariable,
+                                variableProperties[resolvedVariable]!!.owner!!, currentFunction
```
woocash2, line 51: Maybe putting the last argument in a new line would be nicer

app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
```diff
@@ -26,74 +26,63 @@ object VariablePropertiesAnalyzer {
...
                 is Expression.Conditional -> sequenceOf(node.condition, node.resultWhenTrue, node.resultWhenFalse)
-                    .forEach { analyzeVariables(it, currentOwner) }
+                    .forEach { analyzeVariables(it, currentFunction) }
                 is Function -> {
                     node.parameters.forEach {
-                        variableProperties.put(it, VariableProperties(node, false))
+                        variableProperties[it] = VariableProperties(node)
                         // scope of the inner function has not begun yet
-                        it.defaultValue?.let { analyzeVariables(it, currentOwner) }
+                        it.defaultValue?.let { analyzeVariables(it, currentFunction) }
```
woocash2, line 80: image
Using the same name `it` in the outer and the inner lambda may be a bit misleading

woocash2 (changes requested):
You've done mapping from variables to functions, which is fine, but can you also remove `variablesUsedByFunctions` from `FunctionDependenciesAnalyzer` which was initially prepared for this functionality (and `VariableAccessMode` as well)?
Other than that I left a few minor comments.

woocash2 (approved):
Looks cool, approved.

PR #64, opened 2022-11-19
Return typecheck and ictree

ggawryal:


app/src/main/kotlin/compiler/intermediate_form/Addressing.kt
```diff
@@ -0,0 +1,19 @@
+package compiler.intermediate_form
+
+typealias MemoryAddress = ULong
+
+sealed class Addressing {
```
tomekb234, line 5: I think the classes here should extend `Addressing` (same problem for `IntermediateFormTreeNode`).

app/src/main/kotlin/compiler/intermediate_form/IntermediateFormTreeNode.kt
```diff
@@ -1,5 +1,80 @@
...
+    // arithmetic operations work on 32 bit registers
+    data class Add(val resultRegister: Register, val secondRegister: Register) // add reg1, reg2
+    data class AddConst(val resultRegister: Register, val value: Int) // add reg, 42
+    data class AddMem(val resultRegister: Register, val address: Addressing) // add reg, [addr]
+
+    data class Subtract(val resultRegister: Register, val secondRegister: Register)
+    data class SubtractConst(val resultRegister: Register, val value: Int)
+    data class SubtractMem(val resultRegister: Register, val address: Addressing)
+
+    data class Multiply(val resultRegister: Register, val secondRegister: Register)
```
woocash2, line 20: I don't remember whether we were to represent the `asm` instructions precisely here or "atomic" operations in our programming language. Probably the second, but if it was the first then there are a few mismatches with the assembly language. `mul` should only take one param and multiply by `eax`, as well as `div` - this kind of stuff. But yeah, probably the second.

woocash2 (changes requested):
All good, except as Tomek has noticed, the inner classes should extend the outer.

app/src/main/kotlin/compiler/intermediate_form/Addressing.kt
ggawryal: Sure, thanks for the catch

app/src/main/kotlin/compiler/intermediate_form/IntermediateFormTreeNode.kt
ggawryal: I think it's the latter, i.e. nodes should represent atomic operations in our language. I tried to use `x86` like instructions whenever possible, but decided to:
* modify multiplication and division to be consistent with `add/subtract`
* introduce separate nodes for all comparison operators instead of using `test` 
Imho it should faciliate using `IntermediateFormTreeNode` and shouldn't make too much overhead when translating to `asm` later.

woocash2 (approved):

PR #66, opened 2022-11-19
Function control flow

tomekb234:


tomekb234:
The equality checking in tests is a bit messy, as it depends on the fact that the fake `IFTNode.NoOp` is *not* a `data class`. We should eventually review all tests which compare `data class` objects by identity and come up with some common framework for them.

app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
```diff
@@ -12,9 +18,178 @@ object ControlFlow {
...
+            controlFlowGraphs[function] = ControlFlowGraph(
+                treeRoots,
+                entryTreeRoot,
+                unconditionalLinks,
+                conditionalTrueLinks,
+                conditionalFalseLinks
+            )
+        }
+
+        for (global in program.globals) {
```
woocash2, line 182: How about this cool one-liner: `program.globals.filterIsInstance<Program.Global.FunctionDefinition>().forEach { processFunction(it.function) }`

app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
```diff
@@ -12,9 +18,178 @@ object ControlFlow {
...
+                    entryTreeRoot = to
+            }
+
+            fun mapLinkType(list: List<Pair<IFTNode, LinkType>?>, type: LinkType) = list.map { it?.copy(second = type) }
+
+            var last = listOf<Pair<IFTNode, LinkType>?>(null)
+            var breaking: MutableList<Pair<IFTNode, LinkType>?>? = null
+            var continuing: MutableList<Pair<IFTNode, LinkType>?>? = null
+
+            fun processBlock(block: StatementBlock) {
```
woocash2, line 55: Maybe `processStatementBlock` instead?

app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
```diff
@@ -12,9 +18,178 @@ object ControlFlow {
...
+                    if (last.isEmpty())
+                        diagnostics.report(ControlFlowDiagnostic.UnreachableStatement(statement))
+
+                    when (statement) {
+                        is Statement.Evaluation -> addExpression(statement.expression, null)
+
+                        is Statement.VariableDefinition -> {
+                            val variable = statement.variable
+
+                            if (variable.kind != Variable.Kind.CONSTANT && variable.value != null)
```
woocash2, line 86: can we have a `CONSTANT` inside a function's body?

woocash2 (approved):
Looks really nice. Left some minor comments

app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
tomekb234: We can. It would work in the same way as in the global constant, except that the scope would be limited. If constants were only allowed in the global context, they should have had their own variant of `Program.Global`.

app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
tomekb234: Sure.

app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
tomekb234: Renamed.

PR #70, opened 2022-11-20
Add argument resolution

rkilar:


app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+        fun checkFunction(function: Function, global: Boolean) {
+            if (defaultParametersBeforeNonDefault(function)) {
+                diagnostics.report(Diagnostic.ArgumentResolutionError.DefaultParametersNotLast(function))
+            }
+
+            // TODO: generate variables for default parameters
+
+            fun checkBlock(block: StatementBlock) {
+                for (statement in block) {
+                    when (statement) {
```
woocash2, line 164: image
Maybe worth adding `else -> {}` on the bottom

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+                    if (encounteredDefault)
+                        return true
+                } else {
+                    encounteredDefault = true
+                }
+            }
+            return false
+        }
+
+        fun checkFunction(function: Function, global: Boolean) {
```
woocash2, line 155: `global` is not used.

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+        if (variable.value != null) {
+            checkExpression(variable.value)
+        }
+    }
+
+    private fun checkFunctionCalls(program: Program) {
+        fun checkFunction(function: Function) {
+            fun checkBlock(block: StatementBlock) {
+                for (statement in block) {
+                    when (statement) {
```
woocash2, line 115: same as line 164

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+            if (!isMatched[index] && function.parameters[index].defaultValue == null) {
+                diagnostics.report(Diagnostic.ArgumentResolutionError.MissingArgument(functionCall, parameterNames[index]))
+            }
+
+            // TODO: detect default parameter values used in this function call
+        }
+    }
+
+    private fun checkExpression(expression: Expression) {
+        when (expression) {
```
woocash2, line 86: same as line 164

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+
+        val function: Function = nameResolution[functionCall] as Function
+        val parameterNames = function.parameters.map { it.name }
+        val numberOfParameters = parameterNames.size
+        val isMatched = parameterNames.map { false }.toMutableList()
+
+        for ((index, argument) in functionCall.arguments.withIndex()) {
+            if (argument.name == null) {
+                if (index >= numberOfParameters) {
+                    diagnostics.report(Diagnostic.ArgumentResolutionError.TooManyArguments(functionCall))
```
woocash2, line 56: It would be more clear if we checked before the loop if `arguments.size > numberOfParameters`.

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+        }
+    }
+
+    private fun checkVariable(variable: Variable) {
+        if (variable.value != null) {
+            checkExpression(variable.value)
+        }
+    }
+
+    private fun checkFunctionCalls(program: Program) {
```
woocash2, line 111: I'd say the main responsibility of this function is to map arguments to parameters so how about a name `resolveFunctionCallsArguments`?

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+
+            resolver.checkFunctionDefinitions(program)
+            if (resolver.correctDefinitions)
+                resolver.checkFunctionCalls(program)
+
+            return resolver.result
+        }
+    }
+
+    private fun checkFunctionCall(functionCall: Expression.FunctionCall) {
```
woocash2, line 37: similar comment as line 111

woocash2, line 37: similar comment as in line 111

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+                isMatched[parameterIndex] = true
+            }
+        }
+
+        for (index in 0 until numberOfParameters) {
+            if (!isMatched[index] && function.parameters[index].defaultValue == null) {
+                diagnostics.report(Diagnostic.ArgumentResolutionError.MissingArgument(functionCall, parameterNames[index]))
+            }
+
+            // TODO: detect default parameter values used in this function call
```
woocash2, line 81: is this still relevant?

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+            }
+            return false
+        }
+
+        fun checkFunction(function: Function, global: Boolean) {
+            if (defaultParametersBeforeNonDefault(function)) {
+                diagnostics.report(Diagnostic.ArgumentResolutionError.DefaultParametersNotLast(function))
+            }
+
+            // TODO: generate variables for default parameters
```
woocash2, line 160: is this still relevant?

woocash2 (changes requested):
I had one concern with regard to logic, which I commented on in line 81. I also left some optional comments on style and simplicity. Other than that it looks good.

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
rkilar: Tomek wrote in discord about mapping arguments with default values to artificial variables. This feels like a natural place to do it.

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
```diff
@@ -2,17 +2,182 @@ package compiler.semantic_analysis
...
+                }
+            }
+
+            checkBlock(function.body)
+        }
+
+        program.globals.forEach { if (it is Program.Global.FunctionDefinition) checkFunction(it.function) }
+    }
+
+    private fun checkFunctionDefinitions(program: Program) {
```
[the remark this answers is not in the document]
rkilar: The function might do more than that (generate variables for default parameters).

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
rkilar: It might be used while generating variables for default parameters (global functions must have default parameters that are constant).

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
rkilar: You're right about `functionCall.arguments`. I feel like mapping from function calls to default paramters used in that call is needed in later analysis (it uses variables which store default parameter values for local functions).

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
woocash2: Ok, I get it now

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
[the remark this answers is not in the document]
woocash2: So how about renaming every `check*` function to `process*` (since they both compute something and check for errors)?

app/src/main/kotlin/compiler/semantic_analysis/ArgumentResolver.kt
woocash2: Ok, so we can't recur yet, before this `TODO` is completed?

rkilar: We can. I added it to `processExpression`.

woocash2: didn't see, sorry

woocash2 (approved):
Looks cool.
There are still functions with `check*` prefix (line 159, 164) but maybe they're different than the others so follow your preference.
Approved.

PR #78, opened 2022-11-22
Add interfaces related to function details

ignacy123:


app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import compiler.ast.Variable
+
+data class FunctionDetailsGenerator(
+    val depth: Int,
+    val vars: Map<Variable, Boolean>,
+    val parameters: List<Variable>
+)
+
+fun generateCall(args: List<IntermediateFormTreeNode>): ControlFlowGraph {
```
ggawryal, line 11: Shouldn't we also return an IFTNode for function result, outside that CFG: `Pair<ControlFlowGraph, IntermediateFormTreeNode?>`?

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -0,0 +1,30 @@
...
+
+fun genPrologue(): ControlFlowGraph {
+    return TODO()
+}
+
+fun genEpilogue(): ControlFlowGraph {
+    return TODO()
+}
+
+fun genRead(variable: Variable, direct: Boolean): IntermediateFormTreeNode {
```
JHodorHodor, line 23: `isDirect`?

JHodorHodor (approved):
LGTM

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import compiler.ast.Variable
+
+data class FunctionDetailsGenerator(
+    val depth: Int,
+    val vars: Map<Variable, Boolean>,
+    val parameters: List<Variable>
+)
+
+fun generateCall(args: List<IntermediateFormTreeNode>): Pair<ControlFlowGraph, IntermediateFormTreeNode?> {
```
JHodorHodor, line 11: Maybe instead of a pair, we should use an object with named fields?

PR #79, opened 2022-11-23
Restructurise IFTNode

ggawryal:
* Move reading from memory/register/const to leafs
* Other nodes store uses IFTNodes only as args for operation they represent
* Add write to memory and optional write to specific register
* Add LogicalIff and LogicalXor nodes, since it's not possible to create short circuit semantic for them - value of first argument never determines the result


woocash2 (approved):
lgtm

PR #80, opened 2022-11-25
Implement Default Parameter Resolution

JHodorHodor:


app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -29,9 +26,6 @@ class Compiler(val diagnostics: Diagnostics) {
         val parseTree = parser.process(leaves)
 
         val ast = AstFactory.createFromParseTree(parseTree, diagnostics)
-        val nameResolution = NameResolver.calculateNameResolution(ast, diagnostics)
-        val argumentResolution = ArgumentResolver.calculateArgumentToParameterResolution(ast, nameResolution, diagnostics)
-        val expressionTypes = TypeChecker.calculateTypes(ast, nameResolution, argumentResolution, diagnostics)
-        val variableProperties = VariablePropertiesAnalyzer.calculateVariableProperties(ast, nameResolution, diagnostics)
+        val programProperties = Resolver.resolveProgram(ast, diagnostics)
```
JHodorHodor, line 29: I refactored this a little bit to hide some details, let me know what do you think.

app/src/main/kotlin/compiler/intermediate_form/FunctionsUtilities.kt
```diff
@@ -13,15 +13,11 @@ import compiler.common.reference_collections.combineReferenceSets
 import compiler.common.reference_collections.referenceKeys
 import compiler.common.reference_collections.referenceSetOf
 
-object FunctionDependenciesAnalyzer {
+object FunctionsUtilities {
```
JHodorHodor, line 16: I changed this name because as I see we want to pack here some various stuff.

app/src/main/kotlin/compiler/semantic_analysis/DefaultParameterResolver.kt
```diff
@@ -0,0 +1,94 @@
...
+import compiler.ast.Function
+import compiler.ast.NamedNode
+import compiler.ast.Program
+import compiler.ast.Statement
+import compiler.ast.Variable
+import compiler.common.reference_collections.ReferenceHashMap
+import compiler.common.reference_collections.ReferenceMap
+import compiler.common.reference_collections.referenceEntries
+
+object DefaultParameterResolver {
```
JHodorHodor, line 12: Imo, this is the right place to do this resolution.

app/src/test/kotlin/compiler/intermediate_form/FunctionsUtilitiesTest.kt
```diff
@@ -768,4 +768,80 @@ class FunctionDependenciesAnalyzerTest {
 
         assertContentEquals(expectedCallGraph, actualCallGraph)
     }
+
+    @Test fun `test comparison by reference - sets`() {
```
JHodorHodor, line 772: I added one more test to my previous PR (I realized it is missing during coding the current task).

app/src/main/kotlin/compiler/semantic_analysis/Resolver.kt
```diff
@@ -0,0 +1,36 @@
...
+        val variableProperties: ReferenceMap<Any, VariablePropertiesAnalyzer.VariableProperties>,
+        val defaultParameterMapping: ReferenceMap<Function.Parameter, Variable>,
+    )
+
+    fun resolveProgram(ast: Program, diagnostics: Diagnostics): ProgramProperties {
+        val initialNameResolution = NameResolver.calculateNameResolution(ast, diagnostics)
+        val argumentResolution = ArgumentResolver.calculateArgumentToParameterResolution(ast, initialNameResolution, diagnostics)
+        val expressionTypes = TypeChecker.calculateTypes(ast, initialNameResolution, argumentResolution, diagnostics)
+        val variableProperties = VariablePropertiesAnalyzer.calculateVariableProperties(ast, initialNameResolution, diagnostics)
+        val defaultParameterResolutionResult = DefaultParameterResolver.resolveDefaultParameters(ast, initialNameResolution)
```
rkilar, line 26: Shouldn't the variables for default parameter values be included in variable property analysis? They are owned by the outer function/global scope and are accessed whenever the function is called without providing the corresponding argument.

JHodorHodor: I think you're right, I updated the PR.

app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
```diff
@@ -77,8 +85,12 @@ object VariablePropertiesAnalyzer {
                 is Function -> {
                     node.parameters.forEach {
                         variableProperties[it] = VariableProperties(node)
```
JHodorHodor, line 87: @ignacy123 are you sure that this should be (`node`) instead of (`currentFunction`)? Default variables are computed by the parent functions as far as I understand.

app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
```diff
@@ -77,8 +85,12 @@ object VariablePropertiesAnalyzer {
...
-                        // scope of the inner function has not begun yet
-                        it.defaultValue?.let { defValue -> analyzeVariables(defValue, currentFunction) }
+                        if (it.defaultValue != null) {
+                            // scope of the inner function has not begun yet
+                            analyzeVariables(it.defaultValue, currentFunction)
+                            // TODO: uncomment after updating tests
```
JHodorHodor, line 91: I will add tests to this after you approve the shape of the changes (that it should be as Rafał noticed in his comment).

ignacy123 (changes requested):
I like the refactoring of the `Compiler` and `Resolver` classes (IMO it deserves a separate commit at the start of the branch).
I agree that dummy variables for default values should be analyzed in the variable properties analyzer.
Please proceed with the implementation of the missing tests.
My suggestions are only minor and related to naming/readability.

app/src/main/kotlin/compiler/semantic_analysis/DefaultParameterResolver.kt
```diff
@@ -0,0 +1,69 @@
...
+                fun process(vararg bunchOfBlocks: List<Statement>?) = bunchOfBlocks.toList().forEach { block -> block?.forEach { process(it) } }
+
+                when (statement) {
+                    is Statement.Evaluation -> { }
+                    is Statement.VariableDefinition -> { }
+                    is Statement.Assignment -> { }
+                    is Statement.LoopBreak -> { }
+                    is Statement.LoopContinuation -> { }
+                    is Statement.FunctionReturn -> { }
+
```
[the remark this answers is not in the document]
JHodorHodor: Imo, the core purpose of using `when` is to exhaust all possibilities, and in this way to have a compiletime error after changes to the `Statement` class (to be forced to think if some changes are required or not).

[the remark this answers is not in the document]
JHodorHodor: So I'd leave it like it is :thinking:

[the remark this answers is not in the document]
ignacy123: Ok, makes sense to me. Please add a comment with this explanation.

app/src/main/kotlin/compiler/intermediate_form/FunctionsUtilities.kt
JHodorHodor: Probably you're right, I will restore the original name to make merging easier, if we would like to change it later, we can in a separate commit.

app/src/main/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzer.kt
JHodorHodor: Ok

app/src/test/kotlin/compiler/semantic_analysis/ArgumentResolverTest.kt
```diff
@@ -36,6 +42,16 @@ internal class ArgumentResolverTest {
...
+    private fun assertContentEquals(
+        expected: ReferenceMap<Expression.FunctionCall, ReferenceSet<Function.Parameter>>,
+        actual: ReferenceMap<Expression.FunctionCall, ReferenceSet<Function.Parameter>>,
+    ) {
+        expected.referenceKeys.forEach {
+            assertTrue(it in actual.referenceKeys)
+            assertEquals(expected[it]!!.referenceElements, actual[it]!!.referenceElements)
+        }
+    }
+
```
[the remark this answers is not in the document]
JHodorHodor: I think, that it actually should check equality, so I will add a condition. The problem with overridden `equals` lies in `ReferenceSet` being the value type.

app/src/test/kotlin/compiler/semantic_analysis/DefaultParameterResolverTest.kt
```diff
@@ -0,0 +1,251 @@
...
+import compiler.common.reference_collections.referenceKeys
+import compiler.common.reference_collections.referenceMapOf
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotSame
+import kotlin.test.assertTrue
+
+class DefaultParameterResolverTest {
+
+    private fun assertEqualVariables(expected: Variable, actual: Variable) {
```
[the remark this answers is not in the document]
JHodorHodor: Any ideas on how to name this? My best idea is `assertSimilarVariables`.

app/src/test/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzerTest.kt
```diff
@@ -440,4 +477,79 @@ class VariablePropertiesAnalyzerTest {
...
         assertDiagnostics(input, listOf())
     }
+
+    // czynność zewnętrzna() {
+    //     czynność wewnętrzna(x: Liczba = 123) { }
+    //     wewnętrzna(124)
+    // }
+
+    @Test
+    fun `test uncalled default parameter`() {
```
ignacy123, line 487: nit
```suggestion
    fun `test unreferenced default parameter`() {
```
same for below

app/src/test/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzerTest.kt
```diff
@@ -440,4 +477,79 @@ class VariablePropertiesAnalyzerTest {
...
+        val innerFunctionCall = Expression.FunctionCall("wewnętrzna", listOf(Expression.FunctionCall.Argument("x", Expression.NumberLiteral(124))))
+        val function = Function(
+            "zewnętrzna", listOf(), Type.Unit,
+            listOf(
+                Statement.FunctionDefinition(innerFunction),
+                Statement.Evaluation(innerFunctionCall),
+            )
+        )
+        val nameResolution: MutableReferenceMap<Any, NamedNode> = ReferenceHashMap()
+        nameResolution[innerFunctionCall] = innerFunction
```
ignacy123, line 500:
```suggestion
        val nameResolution: MutableReferenceMap<Any, NamedNode> = referenceMapOf(innerFunctionCall to innerFunction)
```
Maybe we should unify all similar calls to this convention as well

app/src/test/kotlin/compiler/semantic_analysis/DefaultParameterResolverTest.kt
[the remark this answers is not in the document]
ignacy123: `assertEqualIgnoreName`?

ignacy123 (approved):

app/src/test/kotlin/compiler/semantic_analysis/ArgumentResolverTest.kt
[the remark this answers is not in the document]
ignacy123: Maybe we override equals for `ReferenceSet` as well? I don't think we'll ever need the default equals, yet we will definitely need to compare `ReferenceSet`s again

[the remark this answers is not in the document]
JHodorHodor: good idea

PR #81, opened 2022-11-26
Function details generator generate call

ggawryal:
I'm not sure if the implementation is correct and sufficient, but it should be close to something good. I can add unit tests after we confirm this is what we wanted and after merging with other branches for other tasks in this sprint.


app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
```diff
@@ -0,0 +1,42 @@
+package compiler.intermediate_form
+
+import compiler.common.reference_collections.ReferenceHashMap
+
+class ControlFlowGraphBuilder {
```
ignacy123, line 5: Would you mind adding some tests for this class? It would be a good way of documenting the properties of the `ControlFlowGraph`s and merging operations performed on them.

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -1,26 +1,77 @@
...
+        for ((linkType, links) in linksToIterateOver) {
+            for ((from, to) in links) {
+                if (to !in finalNodes)
+                    modifiedCfgBuilder.addLink(Pair(from, linkType), to)
+                else {
+                    val newReturnNode = IntermediateFormTreeNode.RegisterWrite(FUNCTION_RESULT_REGISTER, to)
+                    modifiedCfgBuilder.addLink(Pair(from, linkType), newReturnNode)
+                }
+            }
+        }
```
ignacy123, line 58: This should be a function of the `ControlFlowGraph`. Something like `updateFinalLinks`.

ignacy123 (changes requested):
On the face of it, the implementation seems fine, but it definitely needs unit tests.

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -1,26 +1,77 @@
...
 data class FunctionCallIntermediateForm(
     val callGraph: ControlFlowGraph,
     val result: IntermediateFormTreeNode?
 )
 
-fun generateCall(args: List<IntermediateFormTreeNode>): FunctionCallIntermediateForm {
-    return TODO()
+val FUNCTION_RESULT_REGISTER = Register()
+
+fun generateCall(fdg: FunctionDetailsGenerator, args: List<IntermediateFormTreeNode>): FunctionCallIntermediateForm {
```
ignacy123, line 22: Please also move all the functions here to the body of `FunctionDetailsGenerator` and unify the naming convention with [this branch](https://github.com/TCS-Compilers/compiler/blob/9dacb21c97c425af69d3ef695ea3c45cf9977f4f/app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt#L42) before you write tests. This will help with future merging conflicts.

ignacy123 (changes requested):

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val unconditionalLinks = ReferenceHashMap<IFTNode, IFTNode>()
+    private val conditionalTrueLinks = ReferenceHashMap<IFTNode, IFTNode>()
+    private val conditionalFalseLinks = ReferenceHashMap<IFTNode, IFTNode>()
```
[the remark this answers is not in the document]
bebidek: I think there is no possibility of single conditional link (neither true nor false), because in that case we wouldn't know how to continue the execution. In the example above, false link just leads to next instruction (just like the unconditional link at the end of `something`).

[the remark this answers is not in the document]
ggawryal: In  `ControlFlow.kt`, line 95, it seems that such instructions you mentioned will create also conditional `False` link correctly, as @bebidek suggested.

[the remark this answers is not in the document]
ignacy123: Okay, then let's verify that the conditional links have a corresponding pair during `build`

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
ggawryal: Agree, but I'd rather keep `ControlFlowGraph` as simple `data class` with no methods, so I moved this to `ControlFlowGraphBuilder`

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
ggawryal: Moved and added simple unit test. It's hard to test `genCall` without testing `genRead` and `genWrite`, because we cannot extend `FunctionDetailsGenerator` and override these methods to something dummy for such test class. How about using external mocking library (eg. *Mockito*) in our project to faciliate such testing?

ignacy123: That's a great idea. I pushed a commit adding Mockito to the dependencies, you should be able to use it now.

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
```diff
@@ -0,0 +1,70 @@
...
+    fun addLink(from: Pair<IFTNode, CFGLinkType>?, to: IFTNode) {
+        if (from != null) {
+            for (node in listOf(from.first, to))
+                if (!treeRoots.contains(node))
+                    treeRoots.add(node)
+
+            val links = when (from.second) {
+                CFGLinkType.UNCONDITIONAL -> unconditionalLinks
+                CFGLinkType.CONDITIONAL_TRUE -> conditionalTrueLinks
+                CFGLinkType.CONDITIONAL_FALSE -> conditionalFalseLinks
```
ignacy123, line 26: Shouldn't we verify somewhere that these links have corresponding pairs?

ignacy123 (changes requested):
Please:
 - merge the master branch into this one to fix CI
 - add some more detailed tests. I pushed a commit enabling `Mockito`. Try to document where `genCall` is expecting to find the function result so that it doesn't get shadowed in the `genEpilogue` (which will be implemented later as we need to consult it tomorrow)

Otherwise everything looks good, we should merge this soon.

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -1,34 +1,58 @@
...
+    ): FunctionCallIntermediateForm {
+        val cfgBuilder = ControlFlowGraphBuilder()
 
-fun genEpilogue(): ControlFlowGraph {
-    return TODO()
-}
+        var last: Pair<IFTNode, CFGLinkType>? = null
+        // write arg values to params
+        for ((arg, param) in args zip parameters) {
+            val node = genWrite(param, arg, true)
```
ignacy123, line 27: `genCall` will be called from the outer function, so this write might not be direct?

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
```diff
@@ -0,0 +1,70 @@
...
+
+class ControlFlowGraphBuilder(var entryTreeRoot: IFTNode? = null) {
+    private var unconditionalLinks = ReferenceHashMap<IFTNode, IFTNode>()
+    private var conditionalTrueLinks = ReferenceHashMap<IFTNode, IFTNode>()
+    private var conditionalFalseLinks = ReferenceHashMap<IFTNode, IFTNode>()
+    private var treeRoots = ArrayList<IFTNode>()
+
+    init {
+        if (entryTreeRoot != null)
+            treeRoots.add(entryTreeRoot!!)
```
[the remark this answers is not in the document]
ggawryal: Good catch

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
ggawryal: Added verification when calling `build` and it turns out that 2 unit tests (namely `unreachable statement because of continuation` and `function return`) in `FunctionControlFlowTest` are failing - there are true links without matching false links and vice versa. Temporarily I disabled these tests, but we should take a look at them with high priority.

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
ggawryal: Changed to `false`, but still not sure this is correct. Should we add `isDirect` also to `genCall`, or is there any method to decide what to put there?

ggawryal:
* Merged with master and added one more (but sufficient imho) unit test.
* Changed Mockito to Mockk, because Mockito was unable to conviniently mock `FunctionDetailsGenerator` unsing `spy` method, whereas Mockk does it without any problems

ggawryal:
This is a merged branch for both `generateCall` and `genRead/Write` tasks, updated in second sprint for these tasks.
Note, that we've added some FunctionDetailsGenerator fields, so generating `FunctionDetailsGenerator` for all functions will require to be updated - temporarily, I've commented appropriate lines and disabled one test: `FunctionDependenciesAnalyzerTest.test function details generator creation`.

ggawryal:
Merged to one PR with `function-details-full`. Please don't review this to avoid double reviewing.

PR #83, opened 2022-11-27
Function details full

woocash2:
może być tu trochę syfu...

### DIFF
--- a/app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
+++ b/app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
@@ -3,6 +3,21 @@ package compiler.intermediate_form
 import compiler.ast.Function
 import compiler.ast.Type
 import compiler.ast.Variable
+import compiler.common.reference_collections.ReferenceMap
+
+data class FunctionCallIntermediateForm(
+    val callGraph: ControlFlowGraph,
+    val result: IntermediateFormTreeNode?
+)
+
+data class VariableLocation(
+    val depth: Int,
+    val offset: ULong
+)
+
+val FUNCTION_RESULT_REGISTER = Register()
+val STACK_POINTER_REGISTER = Register()
+val BASE_POINTER_REGISTER = Register()
 
 data class FunctionDetailsGenerator(
     val depth: Int,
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -3,83 +3,271 @@ package compiler.intermediate_form
...
+        val savePreviousRbp = IntermediateFormTreeNode.StackPush(
+            IntermediateFormTreeNode.MemoryRead(Addressing.Displacement(displayAddress + memoryUnitSize * depth))
         )
+        val updateRbpAtDepth = IntermediateFormTreeNode.MemoryWrite(
+            Addressing.Displacement(displayAddress + memoryUnitSize * depth),
+            IntermediateFormTreeNode.RegisterRead(BASE_POINTER_REGISTER)
+        )
+        cfgBuilder.addLink(last, savePreviousRbp)
+        last = Pair(savePreviousRbp, CFGLinkType.UNCONDITIONAL)
+        cfgBuilder.addLink(last, updateRbpAtDepth)
```
ignacy123, line 116: This should be a function of `ControlFlowGraphBuilder`. `addLinkToLast` or something similar.

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -3,83 +3,271 @@ package compiler.intermediate_form
...
+        last = Pair(updateRbpAtDepth, CFGLinkType.UNCONDITIONAL)
+
+        // allocate memory and registers for local variables
+        for ((variable, locationType) in variablesLocationTypes.entries) {
+            when (locationType) {
+
```
ignacy123, line 122:
```suggestion

```
nit

ignacy123 (changes requested):
At a glance, the general approach seems okay.
Please add some tests to document the contracts of each method.
I understand this code will undergo some changes after Tuesday, so we'll resume the review process then.

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -3,83 +3,271 @@ package compiler.intermediate_form
...
-fun generateCall(fdg: FunctionDetailsGenerator, args: List<IntermediateFormTreeNode>): FunctionCallIntermediateForm {
-    val cfgBuilder = ControlFlowGraphBuilder()
+data class FunctionDetailsGenerator(
+    val depth: ULong,
+    val variablesLocationTypes: Map<Variable, VariableLocationType>, // should contain parameters
+    val parameters: List<Variable>,
+    val functionCFG: ControlFlowGraph,
+    val function: Function,
+    val variablesDisplayInfo: MutableMap<String, VariableDisplayInfo>,
+    val displayAddress: MemoryAddress
```
rkilar, line 35: I have some questions about constructing `FunctionDetailsGenerator`. `Function.Parameter` are not `Variable`, so the key type should be `Any` or maybe `NamedNode`. Same for `parameters` - it should be `List<Function.Parameter>`. Also what exactly are `variablesDisplayInfo` and `functionCFG`?

woocash2: For each variable we need to know at which level it was defined and what offset from rbp it had then. I assumed we need some info about this in constructor so we can use it in `genRead` and `genWrite`

app/src/main/kotlin/compiler/intermediate_form/FunctionDetailsGenerator.kt
```diff
@@ -3,83 +3,261 @@ package compiler.intermediate_form
...
 val FUNCTION_RESULT_REGISTER = Register()
+val HELPER_REGISTER = Register()
+val BASE_POINTER_REGISTER = Register()
+val STACK_POINTER_REGISTER = Register()
 
-fun generateCall(fdg: FunctionDetailsGenerator, args: List<IntermediateFormTreeNode>): FunctionCallIntermediateForm {
-    val cfgBuilder = ControlFlowGraphBuilder()
+data class FunctionDetailsGenerator(
+    val depth: ULong,
+    val variablesLocationTypes: Map<Variable, VariableLocationType>, // should contain parameters
```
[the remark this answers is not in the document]
woocash2: what u mean?

PR #90, opened 2022-12-02
Add support for global variables in CFG generation.

kamgal00:


app/src/main/kotlin/compiler/common/dfa/VariablesOwner.kt
```diff
@@ -0,0 +1,3 @@
+package compiler.common.dfa
```
bebidek, line 1: I think this file should be in different directory, as it has nothing to do with DFAs.

app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
```diff
@@ -43,6 +45,14 @@ object ControlFlow {
...
         }
 
+        val variableAccessGenerators: ReferenceMap<VariablesOwner, VariableAccessGenerator> = run {
+            val result = ReferenceHashMap<VariablesOwner, VariableAccessGenerator>()
+
+            result.putAll(functionDetailsGenerators)
+            result[VariablePropertiesAnalyzer.GlobalContext] = GlobalVariablesAccessGenerator(variableProperties)
+            result
+        }
+
```
bebidek, line 55: Creating `GlobalVariablesAccessGenerator` object multiple times is not a good idea. Aside from performance concerns, according to Java docs (I couldn't find such information for Kotlin), HashMap "... makes no guarantees as to the order of the map...", which means we could end up with inconsistent offsets in different expressions. I suggest making that object of type `GlobalVariablesAccessGenerator` a parameter of `createGraphForExpression`.

app/src/main/kotlin/compiler/intermediate_form/GlobalVariablesAccessGenerator.kt
```diff
@@ -0,0 +1,35 @@
+package compiler.intermediate_form
+
+import compiler.ast.Variable
+import compiler.common.intermediate_form.VariableAccessGenerator
+import compiler.common.reference_collections.ReferenceHashMap
+import compiler.common.reference_collections.ReferenceMap
+import compiler.semantic_analysis.VariablePropertiesAnalyzer
+
+class GlobalVariablesAccessGenerator(private val variableProperties: ReferenceMap<Any, VariablePropertiesAnalyzer.VariableProperties>) : VariableAccessGenerator {
```
bebidek, line 9:
```suggestion
class GlobalVariablesAccessGenerator(variableProperties: ReferenceMap<Any, VariablePropertiesAnalyzer.VariableProperties>) : VariableAccessGenerator {
```
Parameter `variableProperties` is used solely during initialization, so there is no need to make it a property.

app/src/main/kotlin/compiler/intermediate_form/GlobalVariablesAccessGenerator.kt
```diff
@@ -0,0 +1,35 @@
...
+    companion object {
+        const val VARIABLE_SIZE = 4
+        const val GLOBALS_MEMORY_LABEL = "globals"
+    }
+
+    private val offsets = run {
+        var currentOffset: Long = 0
+        val result = ReferenceHashMap<Variable, Long>()
+
+        result.putAll(variableProperties.filter { it.value.owner === VariablePropertiesAnalyzer.GlobalContext }.map { it.key as Variable to (currentOffset++) * VARIABLE_SIZE }.toMap())
```
bebidek, line 20:
```suggestion
        result.putAll(variableProperties.filter { it.value.owner === VariablePropertiesAnalyzer.GlobalContext }.map { it.key as Variable to (currentOffset++) * VARIABLE_SIZE })
```
`toMap()` is not necessary here, as `putAll` can take list of pairs.

app/src/main/kotlin/compiler/intermediate_form/IntermediateFormTreeNode.kt
```diff
@@ -8,11 +8,12 @@ sealed class IntermediateFormTreeNode {
...
     }
 
-    data class MemoryRead(val address: Addressing) : IntermediateFormTreeNode()
+    data class MemoryRead(val address: IntermediateFormTreeNode) : IntermediateFormTreeNode()
+    data class MemoryLabel(val label: String) : IntermediateFormTreeNode()
     data class RegisterRead(val register: Register) : IntermediateFormTreeNode()
     data class Const(val value: Long) : IntermediateFormTreeNode()
 
-    data class MemoryWrite(val address: MemoryAddress, val node: IntermediateFormTreeNode) : IntermediateFormTreeNode()
+    data class MemoryWrite(val address: IntermediateFormTreeNode, val node: IntermediateFormTreeNode) : IntermediateFormTreeNode()
```
bebidek, line 16: Maybe `node` is not the best name now because first argument is also a node.

app/src/test/kotlin/compiler/semantic_analysis/VariablePropertiesAnalyzerTest.kt
```diff
@@ -72,7 +72,7 @@ class VariablePropertiesAnalyzerTest {
         val input = VariablePropertyInput(Program(listOf(VariableDefinition(variable))), referenceMapOf())
 
         val expectedResults: ReferenceMap<Any, VariableProperties> = referenceMapOf(
-            variable to VariableProperties(null, referenceSetOf(), referenceSetOf()),
+            variable to VariableProperties(VariablePropertiesAnalyzer.GlobalContext, referenceSetOf(), referenceSetOf()),
```
bebidek, line 75: This test should be renamed as now that variable has a parent (assuming parent = owner here).

app/src/test/kotlin/compiler/intermediate_form/ExpressionControlFlowTest.kt
```diff
@@ -324,6 +330,59 @@ class ExpressionControlFlowTest {
...
+        fun cfgForOffset(offset: Long) =
+            IntermediateFormTreeNode.MemoryRead(
+                IntermediateFormTreeNode.Add(
+                    IntermediateFormTreeNode.MemoryLabel("globals"),
+                    IntermediateFormTreeNode.Const(offset)
+                )
+            ).toCfg()
+
+        for (i in 0 until 3) {
+            assertTrue(cfgs.filter { it hasSameStructureAs cfgForOffset(i * 4.toLong()) }.size == 1)
```
bebidek, line 382:
```suggestion
            assertTrue(cfgs.filter { it hasSameStructureAs cfgForOffset(i * 4L) }.size == 1)
```
We can use Long literal.

app/src/test/kotlin/compiler/intermediate_form/ExpressionControlFlowTest.kt
```diff
@@ -324,6 +330,59 @@ class ExpressionControlFlowTest {
...
+        }
+    }
+
```
bebidek, line 388: I don't think we should test details of `GlobalVariablesAccessGenerator` (such as exact name of the label or offset values) in this file. I would suggest making separate test file for this class (which would be relatively short I guess) and in this file just ensure that the proper object is used for global variables (maybe by adding parameter `isGlobal` to `DummyRead/Write`...). That corresponds to my comment about ControlFlow.kt - passing `GlobalVariablesAccessGenerator` as an argument should make testing easier.

bebidek (changes requested):
I see that you managed to add some changes before I finished a review...

app/src/test/kotlin/compiler/intermediate_form/ExpressionControlFlowTest.kt
bebidek: > passing `GlobalVariablesAccessGenerator` as an argument

I wrote it in a little misleading way - of course I meant passing `VariableAccessGenerator` interface.

bebidek (approved):

PR #92, opened 2022-12-03
Fix CFG misconceptions

ignacy123:
This is the result of a merge of the CFGBuilder on the main branch and the one that is used in the function details generator. 
I added the missing tests.

### DIFF
--- a/app/src/main/kotlin/compiler/common/reference_collections/ReferenceHashMap.kt
+++ b/app/src/main/kotlin/compiler/common/reference_collections/ReferenceHashMap.kt
@@ -37,7 +37,7 @@ fun <K, V> combineReferenceMaps(maps: List<ReferenceMap<K, V>>): ReferenceMap<K,
     return combinedMaps
 }
 
-fun <K, V> referenceMapOf(pairs: List<Pair<K, V>>): ReferenceMap<K, V> {
+fun <K, V> referenceHashMapOf(pairs: List<Pair<K, V>>): ReferenceHashMap<K, V> {
     val map = ReferenceHashMap<K, V>()
     for ((key, value) in pairs)
         map[key] = value
@@ -46,6 +46,6 @@ fun <K, V> referenceMapOf(pairs: List<Pair<K, V>>): ReferenceMap<K, V> {
 
 fun <K, V> combineReferenceMaps(vararg maps: ReferenceMap<K, V>): ReferenceMap<K, V> = combineReferenceMaps(maps.asList())
 
-fun <K, V> referenceMapOf(vararg pairs: Pair<K, V>): ReferenceMap<K, V> = referenceMapOf(pairs.asList())
+fun <K, V> referenceHashMapOf(vararg pairs: Pair<K, V>): ReferenceHashMap<K, V> = referenceHashMapOf(pairs.asList())
 
 fun <K, V> ReferenceHashMap<K, V>.copy() = ReferenceHashMap<K, V>().also { it.putAll(this) }
--- a/app/src/test/kotlin/compiler/intermediate_form/ControlFlowGraphBuilderTest.kt
+++ b/app/src/test/kotlin/compiler/intermediate_form/ControlFlowGraphBuilderTest.kt
@@ -1,0 +1,144 @@
+package compiler.intermediate_form
+import compiler.common.reference_collections.referenceHashMapOf
+import kotlin.test.Test
+import kotlin.test.assertEquals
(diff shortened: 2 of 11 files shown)


app/src/test/kotlin/compiler/intermediate_form/ControlFlowGraphBuilderTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Test
+    fun `test makeRoot when already specified`() {
+        val cfgBuilder = ControlFlowGraphBuilder(entryNode)
+        assertFailsWith<IncorrectControlFlowGraphError> { cfgBuilder.makeRoot(entryNode) }
+    }
+
+    @Test
+    fun `test addLink from null sets entryTreeRoot`() {
```
bebidek, line 53: Name doesn't describe the test well. We test overall functionality of `addLink` here, not only its behavior when from = null.

app/src/test/kotlin/compiler/intermediate_form/ControlFlowGraphBuilderTest.kt
```diff
@@ -0,0 +1,144 @@
...
+
+        val cfg = cfgBuilder.build()
+        assertEquals(expectedCFG, cfg)
+    }
+
+    @Test
+    fun `test addAllFrom without modifying entryTreeRoot`() {
+        val cfgBuilder = ControlFlowGraphBuilder(entryNode)
+
+        val remainingTwoNodes = ControlFlowGraphBuilder(secondNode)
```
bebidek, line 95: Actually, this CFG consists of **three** nodes.

app/src/test/kotlin/compiler/intermediate_form/ControlFlowGraphBuilderTest.kt
```diff
@@ -0,0 +1,144 @@
+package compiler.intermediate_form
+import compiler.common.reference_collections.referenceHashMapOf
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
```
bebidek, line 6: Maybe we could test that methods that add link from **all** final trees actually do it. All current tests have only one final tree, so methods linking only one arbitrary tree instead of all would pass the tests.

app/src/main/kotlin/compiler/common/reference_collections/ReferenceHashMap.kt
```diff
@@ -37,7 +37,7 @@ fun <K, V> combineReferenceMaps(maps: List<ReferenceMap<K, V>>): ReferenceMap<K,
     return combinedMaps
 }
 
-fun <K, V> referenceMapOf(pairs: List<Pair<K, V>>): ReferenceMap<K, V> {
+fun <K, V> referenceHashMapOf(pairs: List<Pair<K, V>>): ReferenceHashMap<K, V> {
```
bebidek, line 40: Since we have this function, it would be a good idea to replace usages of `ReferenceHashMap()` constructor with it in the whole code (for the sake of consistency). Also, maybe we should change `referenceSetOf()` into `referenceHashSetOf()` as well.

bebidek (changes requested):

app/src/test/kotlin/compiler/intermediate_form/ControlFlowGraphBuilderTest.kt
ignacy123: https://www.youtube.com/watch?v=iXpxYn2uEEg&ab_channel=EvilBrotherThaddeus

app/src/main/kotlin/compiler/common/reference_collections/ReferenceSet.kt
```diff
@@ -21,17 +21,17 @@ class ReferenceElement<E>(val element: E) {
 val <E> ReferenceSet<E>.referenceElements get() = this.map { ReferenceElement(it) }.toSet()
 
 fun <E> combineReferenceSets(sets: List<ReferenceSet<E>>): ReferenceSet<E> {
-    val combinedSets = ReferenceHashSet<E>()
+    val combinedSets = referenceHashSetOf<E>()
     sets.forEach { combinedSets.addAll(it) }
     return combinedSets
 }
```
bebidek, line 28: To be even cleaner, we could rename this file to `ReferenceHashSet.kt`.

bebidek (approved):

PR #94, opened 2022-12-03
Calculate a unique identifier for every function

ignacy123:
I performed some tests and it turns out that `nasm`'s lexer is incapable of dealing with polish signs (hence some workarounds).


app/src/main/kotlin/compiler/intermediate_form/UniqueIdentifier.kt
```diff
@@ -0,0 +1,37 @@
...
+            throw UnknownCharacter("Unknown character: $character. Only polish signs are allowed outside of the plain ASCII characters.")
+        }
+        val identifierBuilder = StringBuilder()
+        prefix?.let {
+            identifierBuilder.append(it)
+            identifierBuilder.append('|')
+        }
+        curr.forEach { identifierBuilder.append(convertToPlainASCII(it)) }
+        var identifier = identifierBuilder.toString()
+        while (identifier in knownIdentifiers) identifier += "_"
```
bebidek, line 31: That method of resolving conflicts has one issue: after compilation (e.g. in debugger) one wouldn't be able to tell which function is which (technically, that is not true as it's all deterministic, but it would be non-intuitive and probably dependent on order of function declarations).
I would consider alternative scheme in which polish diacritic letter translates to multiple chars (something like 'zażółć' -> 'zaz#o#l#c#').

app/src/test/kotlin/compiler/intermediate_form/UniqueIdentifierTest.kt
```diff
@@ -0,0 +1,51 @@
+package compiler.intermediate_form
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+class UniqueIdentifierTest {
```
bebidek, line 5:
```suggestion
import kotlin.test.assertEquals

class UniqueIdentifierTest {
```
Missing empty line.

app/src/test/kotlin/compiler/intermediate_form/UniqueIdentifierTest.kt
```diff
@@ -0,0 +1,51 @@
...
+        val identifier = factory.build(null, curr)
+
+        assertEquals(curr, identifier.value)
+    }
+
+    @Test
+    fun `test no polish signs in output`() {
+        // nasm lexer is incapable of parsing unicode
+        val factory = UniqueIdentifierFactory()
+        val curr = "polskie_znaki_są_żeś_zaiście_świetneż"
```
bebidek, line 19: I love that test string.

app/src/main/kotlin/compiler/intermediate_form/FunctionDependenciesAnalyzer.kt
```diff
@@ -16,11 +16,33 @@ import compiler.common.reference_collections.referenceSetOf
 import compiler.semantic_analysis.VariablePropertiesAnalyzer
 
 object FunctionDependenciesAnalyzer {
+    fun createUniqueIdentifiers(program: Program): ReferenceMap<Function, UniqueIdentifier> {
+        val uniqueIdentifiers = ReferenceHashMap<Function, UniqueIdentifier>()
+        val identifierFactory = UniqueIdentifierFactory()
+        fun nameFunction(function: Function, pathSoFar: String?): String? {
```
bebidek, line 22:
```suggestion
        fun nameFunction(function: Function, pathSoFar: String?): String {
```
Returned value is never null.

app/src/test/kotlin/compiler/intermediate_form/FunctionDependenciesAnalyzerTest.kt
```diff
@@ -15,6 +15,81 @@ import kotlin.test.Test
...
+        val actualIdentifiers = FunctionDependenciesAnalyzer.createUniqueIdentifiers(program)
+        val expectedIdentifiers = referenceMapOf(
+            inner1 to UniqueIdentifier("some_prefix|zes"),
+            inner2 to UniqueIdentifier("some_prefix|zes_"),
+            inner3 to UniqueIdentifier("some_prefix|zes__"),
+            outerFunction to UniqueIdentifier("some_prefix")
+        )
+        assertEquals(expectedIdentifiers, actualIdentifiers)
+    }
+
```
bebidek, line 92: I have doubts whether these tests should test actual names (which are detail of `UniqueIdentifierFactory` and are tested in `UniqueIdentifierTest.kt`) instead of just whether names are consistent with what `UniqueIdentifierFactory` returns.

app/src/test/kotlin/compiler/intermediate_form/UniqueIdentifierTest.kt
bebidek, line 5: Maybe `UniqueIdentifierFactoryTest` would be a better name, as `UniqueIdentifier` is just a container for a string and obviously we don't test it here (because there is nothing to test).

bebidek (changes requested):
One think that seems missing (but maybe we'll do it somewhere else) is translation of name "główna" into "main", as this is what standard C runtime excepts.

bebidek:
Global variables have just been merged to master, and that PR introduced label 'globals'. Please merge master into your branch and make sure that label won't conflict with e.g. function called 'globals'.

app/src/main/kotlin/compiler/intermediate_form/UniqueIdentifier.kt
```diff
@@ -0,0 +1,37 @@
...
+    fun build(prefix: String?, curr: String): UniqueIdentifier {
+        fun convertToPlainASCII(character: Char): Char {
+            if (character.code in 0..127) return character
+            if (character in knownConversionsToPlainASCII) return knownConversionsToPlainASCII[character]!!
+            throw UnknownCharacter("Unknown character: $character. Only polish signs are allowed outside of the plain ASCII characters.")
+        }
+        val identifierBuilder = StringBuilder()
+        prefix?.let {
+            identifierBuilder.append(it)
+            identifierBuilder.append('|')
```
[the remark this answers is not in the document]
ignacy123: Good catch, it doesn't work. I'll replace it with `$`.

ignacy123:
The identifiers for functions are now prefixed with `fun`, so it is impossible to end up with a `globals` identifier. I also added a couple of tests for that if we ever change the logic (or constants) behind unique identifiers.
I think that the translation of `główna` into `main` identifier is beyond the scope of this issue. I believe we will address this during the next class.

app/src/main/kotlin/compiler/intermediate_form/UniqueIdentifier.kt
```diff
@@ -0,0 +1,72 @@
...
+        public val forbiddenLabels = listOf("globals")
+    }
+
+    private val knownIdentifiers: MutableSet<String> = mutableSetOf()
+    fun build(prefix: String?, curr: String): UniqueIdentifier {
+        fun convertToPlainASCII(character: Char): List<Char> {
+            if (character in charactersAllowedByNasm) return listOf(character)
+            if (character in knownConversionsToAllowedCharacters) return listOf(
+                knownConversionsToAllowedCharacters[character]!!,
+                polishSignSymbol // to make mangled names more readable
```
bebidek, line 41: One thing that bothers me is that we have two diacritic letters ('ż' and 'ź') translating into one ASCII char, so conflicts are possible.

app/src/main/kotlin/compiler/intermediate_form/UniqueIdentifier.kt
```diff
@@ -0,0 +1,72 @@
...
+            throw IllegalCharacter(
+                """
+                Illegal character: $character.
+                Characters allowed by NASM for labels are: ${charactersAllowedByNasm.joinToString()}.
+            """
+            )
+        }
+        val identifierBuilder = StringBuilder().append(prefix ?: functionPrefix).append(levelSeparator)
+        curr.forEach { convertToPlainASCII(it).forEach { character -> identifierBuilder.append(character) } }
+        var identifier = identifierBuilder.toString()
```
bebidek, line 52: This could be 'val'.

app/src/test/kotlin/compiler/intermediate_form/FunctionDependenciesAnalyzerTest.kt
```diff
@@ -14,8 +14,108 @@ import compiler.common.reference_collections.referenceHashSetOf
...
+    @Test fun `test identical function names with accuracy to polish signs do not cause identifier conflict`() {
+        /*
+         czynność some_prefix() {
+             czynność żeś() {}
+             czynność żes() {}
+             czynność zes() {}
```
bebidek, line 77: We should add potential conflict between 'ż' and 'ź' here (reason to be found in another comment).

app/src/test/kotlin/compiler/intermediate_form/UniqueIdentifierFactoryTest.kt
```diff
@@ -0,0 +1,79 @@
+package compiler.intermediate_form
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+class UniqueIdentifierFactoryTest {
```
bebidek, line 6:
```suggestion
import kotlin.test.assertFailsWith

class UniqueIdentifierFactoryTest {
```
Newline missing.

app/src/main/kotlin/compiler/intermediate_form/UniqueIdentifier.kt
```diff
@@ -0,0 +1,72 @@
...
+            'Ó' to 'O',
+            'Ś' to 'S',
+            'Ź' to 'Z',
+            'Ż' to 'Z'
+        )
+        public val forbiddenLabels = listOf("globals")
+    }
+
+    private val knownIdentifiers: MutableSet<String> = mutableSetOf()
+    fun build(prefix: String?, curr: String): UniqueIdentifier {
```
bebidek, line 36: Maybe full 'current' would be clearer (and it is not much longer).

bebidek (changes requested):
Small issue with the naming and some cosmetics.

bebidek (approved):

Issue #95, opened 2022-12-03
Decide which class is responsible for reporting variables used as functions


tomekb234:
And also for reporting functions used as variables. Currently both name resolution and type checker do this.
https://github.com/TCS-Compilers/compiler/pull/53#discussion_r1020980440

tomekb234:
And also for reporting assignments to parameters (which we assume to be immutable). Currently both the type checker and the variable properties analyzer do this.

PR #96, opened 2022-12-03
Tidy up compilation error handling

tomekb234:
The idea is that non-skippable compilation errors (and *only* these errors) are marked with `Compiler.CompilationFailed` class. For different compilation phases there may be different amounts of skippable and non-skippable errors, depending on the invariants for their outputs. The general rule is that if a phase A produces an output which is passed as input to phase B, and B makes *any* assumptions about this input, then A must fulfill this assumption or signal a non-skippable error. The assumptions are usually about the structure of input (as for example the AST builder assumes that a parse tree branch have children corresponding to the production), and about the presence of some mappings (as for example the type checker assumes that all variable assignments are mapped to corresponding declarations). Thus, for example:

- a lexer error is pretty much always skippable - any output sequence of symbols is correct,
- a parser error is only skippable if it does not disturb the structure of the output parse tree,
- a name resolution error is only skippable if it is a name duplication error, because it does not prevent the creation of a full mapping, unlike a "undefined name" error.

For the latter phases I have conservatively assumed that all errors are non-skippable (in particular in the type checker), because we don't quite know yet which output properties will later become assumptions.

An important change is that E2E tests no longer catch arbitrary exceptions - anything thrown which is not `Compiler.CompilationFailed` is assumed to be an internal compiler error.

### DIFF
--- a/app/src/main/kotlin/compiler/ast/AstFactory.kt
+++ b/app/src/main/kotlin/compiler/ast/AstFactory.kt
@@ -1,6 +1,6 @@
 package compiler.ast
 
-import compiler.Compiler.CompilationFailure
+import compiler.Compiler.CompilationFailed
 import compiler.common.diagnostics.Diagnostic
 import compiler.common.diagnostics.Diagnostics
 import compiler.lexer.lexer_grammar.TokenType
@@ -10,7 +10,7 @@ import compiler.parser.grammar.ParserGrammar.Productions
 import compiler.parser.grammar.Symbol
 
 object AstFactory {
-    class AstCreationFailed : CompilationFailure()
+    class AstCreationFailed : CompilationFailed()
 
     // helper methods
     private fun ParseTree<Symbol>.token(): TokenType? = (symbol as? Symbol.Terminal)?.tokenType
@@ -74,19 +74,25 @@ object AstFactory {
         }
     }
 
-    private fun processConst(parseTree: ParseTree<Symbol>): Expression {
+    private fun processConst(parseTree: ParseTree<Symbol>, diagnostics: Diagnostics): Expression {
         val child = (parseTree as ParseTree.Branch).children[0]
 
         return when (child.token()) {
-            TokenType.INTEGER -> try {
-                Expression.NumberLiteral((child as ParseTree.Leaf).content.toInt())
-            } catch (_: NumberFormatException) {
-                throw AstCreationFailed()
-            }
+            TokenType.INTEGER -> Expression.NumberLiteral(
+                (child as ParseTree.Leaf).content.let {
+                    try {
+                        it.toInt()
+                    } catch (_: NumberFormatException) {
+                        diagnostics.report(Diagnostic.InvalidNumberLiteral(it))
+                        0
+                    }
+                }
+            )
 
             TokenType.TRUE_CONSTANT -> Expression.BooleanLiteral(true)
             TokenType.FALSE_CONSTANT -> Expression.BooleanLiteral(false)
             TokenType.UNIT_CONSTANT -> Expression.UnitLiteral
+
             else -> throw IllegalArgumentException()
         }
     }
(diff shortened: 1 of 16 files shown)


app/src/main/kotlin/compiler/ast/AstFactory.kt
```diff
@@ -9,7 +10,7 @@ import compiler.parser.grammar.ParserGrammar.Productions
 import compiler.parser.grammar.Symbol
 
 object AstFactory {
-    class AstCreationFailed : Throwable()
+    class AstCreationFailed : CompilationFailure()
```
bebidek, line 13: It would be cleaner to stick to one convention regarding naming: if we have `CompilationFailure` then it should be `AstCreationFailure`.`
Same with other exception classes from other files.

app/src/test/kotlin/compiler/e2e/VariablePropertiesAnalysisErrorsTest.kt
```diff
@@ -2,13 +2,15 @@ package compiler.e2e
...
-import org.junit.Test
+import kotlin.test.Ignore
+import kotlin.test.Test
 
 class VariablePropertiesAnalysisErrorsTest {
     private fun assertAssignmentToParameterError(program: String) {
         assertErrorOfType(program, Diagnostic.VariablePropertiesError.AssignmentToFunctionParameter::class)
     }
 
+    @Ignore // FIXME: as this is an e2e test, the error is reported by the type checker, and not the variable properties analyzer
```
bebidek, line 13: If we check this in type checker, then why do we even test for that in `VariablePropertiesAnalyzer`?

app/src/test/kotlin/compiler/semantic_analysis/ArgumentResolverTest.kt
```diff
@@ -221,7 +221,9 @@ internal class ArgumentResolverTest {
         val nameResolution = referenceMapOf<Any, NamedNode>()
         val diagnostics = CompilerDiagnostics()
 
-        ArgumentResolver.calculateArgumentToParameterResolution(program, nameResolution, diagnostics)
+        try {
+            ArgumentResolver.calculateArgumentToParameterResolution(program, nameResolution, diagnostics)
+        } catch (_: ArgumentResolver.ResolutionFailed) { }
```
bebidek, line 226: If `ArgumentResolver` is expected to throw an exception, we should verify that the exception is actually thrown. The same comment goes to another test files.

bebidek (commented):
I don't have any direct change requests, just a few suggestions.

app/src/test/kotlin/compiler/e2e/VariablePropertiesAnalysisErrorsTest.kt
tomekb234: I don't know. We'll need to discuss this.
https://github.com/TCS-Compilers/compiler/issues/95

app/src/main/kotlin/compiler/ast/AstFactory.kt
```diff
@@ -75,8 +76,14 @@ object AstFactory {
...
     private fun processConst(parseTree: ParseTree<Symbol>): Expression {
         val child = (parseTree as ParseTree.Branch).children[0]
+
         return when (child.token()) {
-            TokenType.INTEGER -> Expression.NumberLiteral((child as ParseTree.Leaf).content.toInt())
+            TokenType.INTEGER -> try {
+                Expression.NumberLiteral((child as ParseTree.Leaf).content.toInt())
+            } catch (_: NumberFormatException) {
+                throw AstCreationFailed()
+            }
```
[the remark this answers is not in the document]
tomekb234: Done.

app/src/main/kotlin/compiler/ast/AstFactory.kt
tomekb234: Changed `CompilationFailure` to `CompilationFailed`.

app/src/test/kotlin/compiler/semantic_analysis/ArgumentResolverTest.kt
tomekb234: Done.

bebidek (approved):

PR #98, opened 2022-12-04
Add basic instruction set

rkilar:


app/src/main/kotlin/compiler/intermediate_form/Pattern.kt
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    data class BinaryOperator<T : IntermediateFormTreeNode.BinaryOperator>(
+        val type: KClass<T>,
+        val leftPattern: Pattern = AnyNode(),
+        val rightPattern: Pattern = AnyNode()
+    ) : Pattern() {
+        override fun match(node: IntermediateFormTreeNode): Pair<List<IntermediateFormTreeNode>, Map<String, Any>>? {
+            if (!type.isInstance(node)) return null
```
bebidek, line 66: How about:
`val castedNode = type.safeCast(node) ?: return null`
and then using castedNode. Less reflection in the code...
I consider it cleaner, but I guess it's not evidently better.

bebidek (changes requested):

app/src/main/kotlin/compiler/intermediate_form/InstructionSet.kt
```diff
@@ -0,0 +1,211 @@
...
+                inRegisters, outRegister, _ ->
+                listOf(
+                    Instruction.AndRR(inRegisters[0], inRegisters[1]), // AND  reg0, reg1
+                    Instruction.MoveRR(outRegister, inRegisters[0]) //    MOV  out,  reg0
+                )
+            },
+            InstructionPattern(Pattern.BinaryOperator(IntermediateFormTreeNode.BitOr::class)) {
+                inRegisters, outRegister, _ ->
+                listOf(
+                    Instruction.OrRR(inRegisters[0], inRegisters[1]), //  AND  reg0, reg1
```
bebidek, line 105:
```suggestion
                    Instruction.OrRR(inRegisters[0], inRegisters[1]), //  OR  reg0, reg1
```

bebidek (commented):
One small detail. I didn't approved yet because of one issue I mentioned on Discord we need to resolve, but no response so far.

bebidek:
It seems that at least some part of the group agreed on changing `Liczba` to 64 bits. I would suggest that you change that in this PR (as the problem became visible here). It's one simple change in ControlFlow.kt plus one test (according to @tomekb234). Also, we have a conflict with the master. Other than that, PR is ready to get approved.

rkilar:
I found problems with variable size in `ast` and `GlobalVariableAccessGenerator`, but I can't seem to find problems in `ControlFlow.kt`, granted I'm kind of busy and only scanned through it.

bebidek:
You're right, it was in ast, not in ControlFlow. Sorry for confusion.

app/src/test/kotlin/compiler/e2e/common/E2eAsserter.kt
```diff
@@ -20,6 +20,7 @@ object E2eAsserter {
 
     fun <T : Diagnostic> assertProgramGeneratesDiagnostics(program: String, diagnostics: List<Diagnostic>, diagnosticType: KClass<T>) {
         val actualDiagnostics = runProgram(program)
+        println(actualDiagnostics.toList())
```
bebidek, line 23: I am pretty sure this line was left here by mistake.

app/src/main/kotlin/compiler/intermediate_form/IntermediateFormTreeNode.kt
```diff
@@ -45,9 +45,12 @@ sealed class IntermediateFormTreeNode {
     data class GreaterThan(override val left: IntermediateFormTreeNode, override val right: IntermediateFormTreeNode) : BinaryOperator(left, right)
     data class GreaterThanOrEquals(override val left: IntermediateFormTreeNode, override val right: IntermediateFormTreeNode) : BinaryOperator(left, right)
 
-    data class StackPush(val Node: IntermediateFormTreeNode) : IntermediateFormTreeNode()
+    data class StackPush(val node: IntermediateFormTreeNode) : IntermediateFormTreeNode()
     class StackPop : IntermediateFormTreeNode()
 
+    data class Call(val address: IntermediateFormTreeNode) : IntermediateFormTreeNode()
+    class Return : IntermediateFormTreeNode()
```
bebidek, line 52: We decided that there will be no `Return` IFTNode.

bebidek (approved):

bebidek:
I made these two changes, if you accept them (after all, it's your PR), you can merge.

PR #100, opened 2022-12-04
Unify error reporting

JHodorHodor:


app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -36,6 +36,8 @@ class Compiler(val diagnostics: Diagnostics) {
             val programProperties = Resolver.resolveProgram(ast, diagnostics)
 
             // TODO: generate the code
-        } catch (_: CompilationFailed) { }
+        } catch (_: CompilationFailed) { } finally {
+            println(diagnostics)
```
tomekb234, line 40: This doesn't have to be in a `finally` block - if there is an internal error, then we are in trouble anyway.

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -1,165 +1,292 @@
...
-        }
-
-        data class ParameterAssignment(val assignment: Statement.Assignment, val parameter: Function.Parameter) : TypeCheckingError() {
-            override fun toString() = "Cannot assign to a parameter"
-        }
-
-        data class FunctionAssignment(val assignment: Statement.Assignment, val function: Function) : TypeCheckingError() {
-            override fun toString() = "Cannot assign to a function"
-        }
+    sealed class ResolutionError(astNodes: List<AstNode>) : Diagnostic {
```
tomekb234, line 55: The argument probably would be more convenient as `vararg AstNode`.

app/src/main/kotlin/compiler/parser/Parser.kt
```diff
@@ -148,26 +148,22 @@ class Parser<S : Comparable<S>>(
...
+                        callResult = ParseTree.Branch(location, symbol, parsedSubtrees, action.production)
                         callStack.removeLast()
                     }
 
                     null -> {
                         // Report a parsing error when the parsing table does not tell what to do.
-                        val expectedSymbols = parseActions.keys.filter { it.first == dfa && it.second == state }.mapNotNull { it.third }
-                        diagnostics.report(Diagnostic.ParserError(lookahead?.symbol, lookaheadStart, lookaheadEnd, expectedSymbols))
+                        // val expectedSymbols = parseActions.keys.filter { it.first == dfa && it.second == state }.mapNotNull { it.third }
+                        diagnostics.report(Diagnostic.ParserError.UnexpectedToken(lookahead?.symbol, LocationRange(lookaheadStart, lookaheadEnd), /*expectedSymbols*/))
```
tomekb234, line 166: Why no longer telling the expected symbols?

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -1,165 +1,292 @@
...
+                override val errorMessage = "The parameter is owned by the function ${owner.name} and is assigned in the function ${assignedIn.name}."
+            }
         }
 
-        data class ContinuationOutsideOfLoop(val loopContinuation: Statement.LoopContinuation) : ControlFlowDiagnostic() {
-            override fun isError() = true
-            override fun toString() = "Cannot use 'continue' outside of loop"
+        sealed class ControlFlowDiagnostic(astNodes: List<AstNode>) : ResolutionError(astNodes) {
+
+            sealed class Warnings(astNodes: List<AstNode>) : ControlFlowDiagnostic(astNodes) {
```
bebidek, line 265: It's a little weird that a class representing warnings is derived from `ResolutionError` class.

bebidek (commented):
@tomekb234 highlighted most important points, I just add some minor suggestions.

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
JHodorHodor: I tried and I had a problem with passing it to the child class (some typing error).

app/src/main/kotlin/compiler/parser/Parser.kt
JHodorHodor: Usually, the list of expected symbols is very long and not user-friendly. Some EXPR512 and others. This has no meaning if you don't know the grammar details. Maybe, we can add a task to tidy this up.

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
JHodorHodor: `ResolutionDiagnostic`?

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -1,165 +1,292 @@
...
 
-        data class FunctionAsValue(val expression: Expression.Variable, val function: Function) : TypeCheckingError() {
-            override fun toString() = "Cannot use a function as a value"
-        }
+        data class ObjectAssociatedToError(val message: String, val location: LocationRange?)
 
-        data class VariableCall(val call: Expression.FunctionCall, val variable: Variable) : TypeCheckingError() {
-            override fun toString() = "Cannot call a variable"
-        }
+        val associatedObjects = astNodes.map { ObjectAssociatedToError(it.print(), it.location) }
```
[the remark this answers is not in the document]
tomekb234: On the other hand, displaying something differently than what the user has *exactly* written may be confusing. And as for the long statements (for example conditionals), we don't have to use the location range spanning the entire statement, but instead only its parts, such as `if (...)` without the following block - these locations could be extracted from parse tree and passed as additional locations. But I suppose your solution also kind of works, so let's leave this issue for now.

app/src/main/kotlin/compiler/parser/Parser.kt
tomekb234: Perhapse a good idea would be to display the *tokens* that these non-terminal symbols could begin with (the FIRST set from the grammar analysis).

bebidek: Is this conversation resolved?

bebidek: I assume that we stay with what we have right now, we'll surely discuss that matter with whole team.

app/src/test/kotlin/compiler/semantic_analysis/Common.kt
```diff
@@ -0,0 +1,18 @@
+package compiler.semantic_analysis
+import compiler.common.diagnostics.Diagnostic
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+fun assertResolutionErrorsEquals(expected: List<Diagnostic>, actual: List<Diagnostic>) {
```
bebidek, line 6: I would change name of that method too (in fact, this very name made me notice the naming inconsistency about `ResolutionDiagnostic`).

bebidek (changes requested):
One renaming suggestion, and there are some conflicts with master to solve.

app/src/main/kotlin/compiler/parser/Parser.kt
JHodorHodor: I think, that Tomek's idea is very nice, also sounds simple, so maybe I can implement it within this PR.

app/src/main/kotlin/compiler/ast/AstNode.kt
```diff
@@ -4,4 +4,78 @@ import compiler.lexer.LocationRange
...
+                    is Expression.UnaryOperation -> "${this.kind} ${this.operand}"
+                    is Expression.UnitLiteral -> "nic"
+                    is Expression.Variable -> this.name
+                }
+
+                is Variable -> "${this.kind} ${this.name}: ${this.type}${if (this.value != null) "= ${this.value.printSimple()}" else ""}"
+                is Function -> "czynność ${this.name}(${this.parameters.joinToString { it.printSimple() }}) -> ${this.returnType}"
+                is Function.Parameter -> "${this.name}: ${this.type}${if (this.defaultValue != null) " = ${this.defaultValue.printSimple()}" else ""}"
+                is Expression.FunctionCall.Argument -> "${if (this.name != null) "${this.name} = " else ""}${this.value.printSimple()}"
+                else -> ""
```
bebidek, line 31: Do we want this `else` here? It is not needed now (as all possibilities are listed) and if we add new expressions in the future, we probably would like Kotlin compiler to remind us that we need to update this method as well.

app/src/main/kotlin/compiler/ast/AstNode.kt
```diff
@@ -4,4 +4,78 @@ import compiler.lexer.LocationRange
...
+                    is Statement.LoopBreak -> "przerwij"
+                    is Statement.LoopContinuation -> "pomiń"
+                    is Statement.VariableDefinition -> "definition of << ${this.variable.printSimple()} >>"
+                }
+
+                is Variable -> "variable << ${this.printSimple()} >>"
+                is Function -> "function << ${this.printSimple()} >>"
+                is Function.Parameter -> "function parameter << ${this.printSimple()} >>"
+                is Expression.FunctionCall.Argument -> "function argument << ${this.printSimple()} >>"
+                else -> ""
```
bebidek, line 75: Same as above.

bebidek (changes requested):

app/src/main/kotlin/compiler/ast/AstNode.kt
JHodorHodor: Oh, I forgot... I had a problem with local compilation, the compiler was complaining that "when is not exhaustive" (very weird, especially since on git, the compilation was fine), and I wanted to compile the project to test it. Then, I forgot to change this back.

app/src/main/kotlin/compiler/ast/AstNode.kt
```diff
@@ -4,4 +4,78 @@ import compiler.lexer.LocationRange
...
+                is Function.Parameter -> "${this.name}: ${this.type}${if (this.defaultValue != null) " = ${this.defaultValue.printSimple()}" else ""}"
+                is Expression.FunctionCall.Argument -> "${if (this.name != null) "${this.name} = " else ""}${this.value.printSimple()}"
+                else -> ""
+            }
+        )
+
+        return stringBuilder.toString()
+    }
+
+    fun print(): String {
```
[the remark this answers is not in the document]
JHodorHodor: True, do you like the names now?

app/src/main/kotlin/compiler/common/diagnostics/Diagnostic.kt
```diff
@@ -1,165 +1,296 @@
...
+
+        override fun toString(): String = "Parser Error\n${errorMessage}\n"
+
+        class UnexpectedToken(
+            symbol: Any?,
+            location: LocationRange,
+            expectedSymbols: List<Any?>,
+        ) : ParserError() {
+            override val errorMessage = StringBuilder().apply {
+                if (symbol != null) append("The symbol <<${symbol.toString().drop(1)}>> is unexpected at location $location.")
```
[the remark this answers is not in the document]
JHodorHodor: Good idea

app/src/main/kotlin/compiler/ast/AstNode.kt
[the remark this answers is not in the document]
bebidek: Yes, I think they're much better now.

bebidek (approved):

PR #102, opened 2022-12-06
Fix a potential bug with CFGs

tomekb234:


app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraph.kt
```diff
@@ -11,8 +11,14 @@ data class ControlFlowGraph(
     val conditionalTrueLinks: ReferenceMap<IFTNode, IFTNode>,
     val conditionalFalseLinks: ReferenceMap<IFTNode, IFTNode>
 ) {
-    val finalTreeRoots: List<IFTNode> get() = treeRoots.filter {
-        it !in unconditionalLinks && it !in conditionalTrueLinks && it !in conditionalFalseLinks
+    val finalTreeRoots: List<Pair<IFTNode, CFGLinkType>> get() = treeRoots.filter {
```
JHodorHodor, line 14: Can we create a class e.g. `OutLink`, with `IFTNode` and `CFGLinkType`?

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
```diff
@@ -43,7 +50,15 @@ class ControlFlowGraphBuilder(@JvmField var entryTreeRoot: IFTNode? = null) {
     }
 
     fun addLinksFromAllFinalRoots(linkType: CFGLinkType, to: IFTNode) {
-        finalTreeRoots.forEach { addLink(Pair(it, linkType), to) }
+        finalTreeRoots.forEach {
+            if (it.second == CFGLinkType.UNCONDITIONAL)
+                addLink(it.copy(second = linkType), to)
```
JHodorHodor, line 55: I think that `addLink(it.first to linkType, to)` is more readable

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraph.kt
JHodorHodor, line 14: IntelliJ says that this could be private. Also the name is no longer valid, maybe. `outLinks`, or sth would be better.

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
```diff
@@ -9,8 +9,15 @@ class ControlFlowGraphBuilder(@JvmField var entryTreeRoot: IFTNode? = null) {
...
-        it !in unconditionalLinks && it !in conditionalTrueLinks && it !in conditionalFalseLinks
+
+    val finalTreeRoots: List<Pair<IFTNode, CFGLinkType>> get() = treeRoots.filter {
+        it !in unconditionalLinks && (it !in conditionalTrueLinks || it !in conditionalFalseLinks)
+    }.map {
+        it to when (it) {
+            in conditionalTrueLinks -> CFGLinkType.CONDITIONAL_FALSE
+            in conditionalFalseLinks -> CFGLinkType.CONDITIONAL_TRUE
+            else -> CFGLinkType.UNCONDITIONAL
+        }
```
JHodorHodor, line 20:
```suggestion
    val finalTreeRoots: List<Pair<IFTNode, CFGLinkType>> = build().finalTreeRoots
```
?

JHodorHodor: Actually, in some places in this file it is used as `build().finalTreeRoots`, is there a difference?

JHodorHodor (changes requested):
Wygląda dobrze, dodałem pare sugesti stylistycznych.

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraph.kt
tomekb234: This type (`List<Pair<IFTNode, CFGLinkType>>`) is also used by control flow generators, so this change would make the property inconvenient to use.

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
tomekb234: Idk, I haven't originally written this code. This property was copy-pasted from `ControlFlowGraph` and so I changed the copy.

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraphBuilder.kt
tomekb234: Changed to `Pair(it.first, linkType)`, because the two meanings of `to` may be confusing.

app/src/main/kotlin/compiler/intermediate_form/ControlFlowGraph.kt
JHodorHodor: What I meant is that now `finalTreeRoots` is not a collection of nodes but some more complex structure.

JHodorHodor: Yes, but in control flow generators also List<OutLink> would look better. But perhaps too much effort now, let's leave it.

JHodorHodor (approved):

PR #104, opened 2022-12-06
Template

JHodorHodor:


app/src/main/kotlin/compiler/intermediate_form/Instruction.kt
```diff
@@ -1,42 +1,56 @@
...
-    data class SetLtEqR(val reg: Register) : Instruction() //                   SETLE reg
-    data class SetGtR(val reg: Register) : Instruction() //                     SETG  reg
-    data class SetGtEqR(val reg: Register) : Instruction() //                   SETGE reg
-
-    data class CallR(val reg: Register) : Instruction() //                      CALL reg
-    data class CallM(val address: Addressing) : Instruction() //                CALL mem
-    class Ret : Instruction() //                                                RET
+sealed class Instruction : AsmAble {
+
+    abstract class ConditionalJumpInstruction : Instruction() {
```
JHodorHodor, line 5: I'm not sure about this. Is this right?

app/src/main/kotlin/compiler/intermediate_form/AsmAble.kt
```diff
@@ -0,0 +1,3 @@
+package compiler.intermediate_form
+
+sealed interface AsmAble
```
tomekb234, line 3: `Asmable` (this is supposed to be a single adjective)

app/src/main/kotlin/compiler/intermediate_form/Instruction.kt
tomekb234: What do we need these `abstract class`es for?

JHodorHodor: Dr Herman said that it will be needed in liveness analysis. To know if an instruction is a jump (and what king of jump) or no. So now, we can do "when".

app/src/main/kotlin/compiler/intermediate_form/Liveliness.kt
```diff
@@ -0,0 +1,10 @@
+package compiler.intermediate_form
+
+object Liveliness {
```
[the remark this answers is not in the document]
JHodorHodor: Yes, probably you're right. IntelliJ suggested this.

app/src/main/kotlin/compiler/intermediate_form/AsmAble.kt
JHodorHodor: True

app/src/main/kotlin/compiler/intermediate_form/Instruction.kt
tomekb234: So, shouldn't these be `sealed class`es?

JHodorHodor: Oh, that's true!

tomekb234: ...and the inner classes should extend the outer instead of `Instruction`.

JHodorHodor: Done

tomekb234: And shouldn't there be cases for `ConditionalJumpInstruction`? Or is someone else responsible for adding them?

JHodorHodor: I don't see any among the existing instructions, so probably someone responsible for the Instruction Set will add them.

tomekb234 (approved):

app/src/main/kotlin/compiler/intermediate_form/Instruction.kt
```diff
@@ -1,42 +1,56 @@
...
+        abstract val targetLabel: String
+    }
+
+    abstract class UnconditionalJumpInstruction : Instruction() {
+        abstract val targetLabel: String
+
+        data class CallR(val reg: Register) : Instruction() //                      CALL reg
+        data class CallM(val address: Addressing) : Instruction() //                CALL mem
```
bebidek: I don't think we should treat these as jump instructions. From the perspective of single function it is `InPlaceInstruction`.

PR #106, opened 2022-12-08
Fix handling function parameters in ExpressionControlFlow

bebidek:


app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
```diff
@@ -128,9 +128,9 @@ object ControlFlow {
             addNextCFG(ControlFlowGraphBuilder().apply { addLink(null, nextTree) }.build())
         }
 
-        fun makeVariableReadNode(variable: Variable): IntermediateFormTreeNode {
-            val owner = variableProperties[variable]!!.owner
-            return variableAccessGenerators[owner]!!.genRead(variable, owner == currentFunction)
+        fun makeReadNode(namedNode: NamedNode): IntermediateFormTreeNode {
```
JHodorHodor, line 131: Is _a function_ allowed here? If this is either variable or param, maybe instead of `namedNode`, we should call the argument `variableOrParam`, or sth similar. Obviously, the best solution would be to create a superclass for `Variable` and `Parameter` and also properly type `variableProperties`, but that may require some work.

JHodorHodor (commented):
Dodałem jeden komentarz, który jest raczej opcjonalny, ale można się zastanowić.

app/src/main/kotlin/compiler/intermediate_form/ControlFlow.kt
bebidek: How about `readableNode`?

JHodorHodor: Sounds ok, maybe we can even leave namedNode, but add a comment that the argument shouldn't be a function

bebidek: Couldn't decide, so I did both...

PR #113, opened 2022-12-10
Liveness analysis

bebidek:
Some changes in Instruction.kt are temporary and are meant to be overwritten by "Instructions" PR.

### DIFF
--- a/app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
+++ b/app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
@@ -21,7 +21,7 @@ abstract class DataFlowAnalyser<SL> {
         val instructionList = linearProgram.filterIsInstance<Instruction>()
 
         // calculate label to instruction translation
-        val labelToInstruction = referenceHashMapOf<String, Instruction>()
+        val labelToInstruction = mutableMapOf<String, Instruction>()
         for ((index, asmable) in linearProgram.withIndex().reversed())
             if (asmable is Label)
                 labelToInstruction[asmable.label] = linearProgram[index + 1].let {
--- a/app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
+++ b/app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
@@ -33,7 +33,7 @@ abstract class DataFlowAnalyser<SL> {
 
         // calculate the predecessors map (forward direction)
         val predecessors = referenceHashMapOf<Instruction, MutableList<Instruction>>()
-        linearProgram.filterIsInstance<Instruction>().forEach { predecessors[it] = mutableListOf() }
+        instructionList.forEach { predecessors[it] = mutableListOf() }
         fun addPredecessor(from: Instruction, to: Instruction) {
             if (backward) predecessors[from]!!.add(to)
             else predecessors[to]!!.add(from)
--- a/app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
+++ b/app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
@@ -1,0 +1,110 @@
+package compiler.intermediate_form
+
+import compiler.common.reference_collections.ReferenceMap
+import compiler.common.reference_collections.ReferenceSet
+import compiler.common.reference_collections.referenceHashMapOf
+import compiler.common.reference_collections.referenceHashSetOf
+import compiler.common.reference_collections.referenceMapOf
+
+abstract class DataFlowAnalyser<SL> {
+    abstract val backward: Boolean
+
+    abstract val entryPointValue: SL
+
+    abstract val latticeMaxElement: SL
+
+    abstract fun latticeMeetFunction(elements: Collection<SL>): SL
+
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        return referenceMapOf(predecessors.map { it.toPair() })
+    }
+
+    private fun getEntryPoints(instructionList: List<Instruction>): ReferenceSet<Instruction> {
+        return if (backward)
+            referenceHashSetOf(instructionList.filterIsInstance<Instruction.RetInstruction>())
+        else
+            referenceHashSetOf(instructionList.first())
```
JHodorHodor, line 66: Is `instructionList` empty for the empty program?

app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
```diff
@@ -0,0 +1,110 @@
+package compiler.intermediate_form
+
+import compiler.common.reference_collections.ReferenceMap
+import compiler.common.reference_collections.ReferenceSet
+import compiler.common.reference_collections.referenceHashMapOf
+import compiler.common.reference_collections.referenceHashSetOf
+import compiler.common.reference_collections.referenceMapOf
+
+abstract class DataFlowAnalyser<SL> {
```
JHodorHodor, line 9:
```suggestion
abstract class DataFlowAnalyzer<SL> {
```
let's be consistent ;)

app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
```diff
@@ -0,0 +1,110 @@
...
+    private fun getEntryPoints(instructionList: List<Instruction>): ReferenceSet<Instruction> {
+        return if (backward)
+            referenceHashSetOf(instructionList.filterIsInstance<Instruction.RetInstruction>())
+        else
+            referenceHashSetOf(instructionList.first())
+    }
+
+    data class DataFlowResult<SL>(val inValues: ReferenceMap<Instruction, SL>, val outValues: ReferenceMap<Instruction, SL>)
+
+    fun analyse(linearProgram: List<Asmable>): DataFlowResult<SL> {
```
JHodorHodor, line 71:
```suggestion
    fun analyze(linearProgram: List<Asmable>): DataFlowResult<SL> {
```

app/src/main/kotlin/compiler/intermediate_form/Liveness.kt
```diff
@@ -1,10 +1,65 @@
...
+
+        // run analysis and fill the graphs
+        val dataFlowResult = LivenessDataFlowAnalyser.analyse(linearProgram)
+        val outLiveRegisters = dataFlowResult.outValues.mapValues { (instr, regs) -> combineReferenceSets(regs, referenceHashSetOf(instr.regsDefined.toList())) }
+
+        instructionList.forEach {
+            for (definedReg in it.regsDefined)
+                for (regLiveOnOutput in outLiveRegisters[it]!!)
+                    if (definedReg !== regLiveOnOutput) {
+                        if (it is Instruction.InPlaceInstruction.MoveRR && it.reg_dest == definedReg && it.reg_src == regLiveOnOutput) {
```
JHodorHodor, line 53: I'm not 100% sure, but if the instruction is `x := y`, and `y` is never used again, then y is not in `outLiveRegisters[it]!!`, and thus there is no `x--y` edge in the copy graph, but we would like it to be there.

app/src/test/kotlin/compiler/intermediate_form/LivenessTest.kt
```diff
@@ -0,0 +1,275 @@
...
+        val livenessGraphs = Liveness.computeLiveness(linearProgram)
+        assertEquals(expectedLivenessGraphs, livenessGraphs)
+    }
+
+    @Test
+    fun `copying and interference`() {
+        val linearProgram = listOf(
+            Instruction.InPlaceInstruction.Dummy(regsDefined = listOf(reg1, reg2, reg3)),
+            Instruction.InPlaceInstruction.MoveRR(reg2, reg1),
+            Instruction.InPlaceInstruction.Dummy(regsUsed = listOf(reg1, reg2, reg3)),
```
JHodorHodor, line 244: According to my comment above. If I understand correctly, the removal of `reg1` in this line shouldn't change the result and it changes.

JHodorHodor (changes requested):
Mam pewną wątpliwość, czy program działa tak jak powinien, zostawiłem w komentarzach.

app/src/main/kotlin/compiler/intermediate_form/DataFlowAnalyser.kt
bebidek: I assume that every execution path must lead to RET instruction.

app/src/main/kotlin/compiler/intermediate_form/Liveness.kt
bebidek: You're right, corrected and added a test for that.

JHodorHodor (approved):

PR #114, opened 2022-12-11
Implement CFG Linearization

tomekb234:


app/src/main/kotlin/compiler/intermediate_form/Linearization.kt
```diff
@@ -1,10 +1,111 @@
...
+
+                if (target in labels)
+                    addJump(labels[target]!!)
+                else
+                    dfs(target, nextLabel)
+            } else if (node in cfg.conditionalTrueLinks && node in cfg.conditionalFalseLinks) {
+                val targetWhenTrue = cfg.conditionalTrueLinks[node]!!
+                val targetWhenFalse = cfg.conditionalFalseLinks[node]!!
+
+                if (targetWhenTrue in labels) {
```
JHodorHodor, line 62: A matter of taste, but I would divide this into 4 conditions to improve readability.

app/src/test/kotlin/compiler/intermediate_form/LinearizationTest.kt
```diff
@@ -0,0 +1,437 @@
...
+        }
+
+        override fun coverConditional(iftNode: IntermediateFormTreeNode, targetLabel: String, invert: Boolean): List<Instruction> {
+            val instruction = Instruction.NoOp()
+            conditional[instruction] = Triple(iftNode, !invert, targetLabel)
+            return listOf(instruction)
+        }
+    }
+
+    private sealed class AsmablePattern
```
JHodorHodor, line 30: Maybe `AsmablePatternTest`?

app/src/test/kotlin/compiler/intermediate_form/LinearizationTest.kt
```diff
@@ -0,0 +1,437 @@
...
+
+        assertLinearizationMatches(
+            listOf(
+                NextLabel,
+                Conditional(node, false, 0)
+            ),
+            result
+        )
+    }
+}
```
JHodorHodor, line 437: Tests are pretty exhaustive but maybe we could add something with nested loops or conditions?

JHodorHodor (changes requested):
Kilka stylistycznych uwag i jedna większw wątpliwość (czy przypadkiem `endLabel` nie powinno być zamienione z `nextLabel`).

app/src/main/kotlin/compiler/intermediate_form/Linearization.kt
```diff
@@ -1,10 +1,111 @@
...
+        }
+
+        fun dfs(node: IFTNode, nextLabel: String) {
+            addLabel(labels[node] ?: assignLabel(node))
+
+            fun addUncoditional() {
+                instructions.addAll(covering.coverUnconditional(node))
+            }
+
+            fun addConditional(whenTrue: Boolean, targetLabel: String) {
```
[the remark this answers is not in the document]
tomekb234: We use "link type" in the context of CFGs to mean either "unconditional", "conditional when the result is true", or "conditional when the result false". Since the name of the function is already `addConditional`, changing the parameter name to `linkType` would be confusing.

app/src/test/kotlin/compiler/intermediate_form/LinearizationTest.kt
tomekb234: We don't test patterns, but use patterns for testing.

app/src/test/kotlin/compiler/intermediate_form/LinearizationTest.kt
```diff
@@ -0,0 +1,437 @@
...
+        override fun coverConditional(iftNode: IntermediateFormTreeNode, targetLabel: String, invert: Boolean): List<Instruction> {
+            val instruction = Instruction.NoOp()
+            conditional[instruction] = Triple(iftNode, !invert, targetLabel)
+            return listOf(instruction)
+        }
+    }
+
+    private sealed class AsmablePattern
+
+    private class Uncoditional(val node: IFTNode) : AsmablePattern()
```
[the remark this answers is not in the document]
tomekb234: Fixed.

app/src/test/kotlin/compiler/intermediate_form/LinearizationTest.kt
tomekb234: Added two more tests. (Note that the tests had already covered all lines in `Linearization`.)

app/src/main/kotlin/compiler/intermediate_form/Linearization.kt
```diff
@@ -1,10 +1,111 @@
...
+                    dfs(targetWhenFalse, nextLabel)
+                }
+            } else if (node in cfg.conditionalTrueLinks) {
+                val target = cfg.conditionalTrueLinks[node]!!
+
+                if (target in labels) {
+                    addConditional(true, labels[target]!!)
+                    addJump(endLabel)
+                } else {
+                    addConditional(false, endLabel)
```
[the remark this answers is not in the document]
tomekb234: If there is only a link for the `true` case, then in the `false` case the control flow reaches the end. The `nextLabel` is only used to skip unnecessary jumps (see how `addJump` works), and here a jump to `endLabel` is necessary, because it is followed by some sequence of instructions generated by `dfs` in the following line.

By the way, this situation will never occur in practice, because all CFGs generated for functions are suffixed by epilogues (`ControlFlow.attachPrologueAndEpilogue`), and so there won't be any node with only one conditional link (because the are no conditionals in the epilogue). This code, however, does not assume this, and covers this situation anyway. (And so it is possible to linearize the function's CFG and its epilogue's CFG separately.)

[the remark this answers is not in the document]
JHodorHodor: Thanks for explanation, it's clear now.

JHodorHodor (approved):

PR #115, opened 2022-12-11
Extend the instruction set

rkilar:
I renamed some classes to avoid further confusion.

### DIFF
--- a/app/src/main/kotlin/compiler/intermediate_form/IFTPattern.kt
+++ b/app/src/main/kotlin/compiler/intermediate_form/IFTPattern.kt
@@ -7,10 +7,12 @@ import kotlin.reflect.safeCast
 // Not to be confused with the Pattern interface
 sealed class IFTPattern {
 
+    data class MatchResult(val subtrees: List<IntermediateFormTreeNode>, val args: Map<String, Any>)
+
     // Returns null if the pattern doesn't match provided intermediate form tree
     // If the tree matches, it returns the list of unmatched subtrees and
     // captured values (mainly used to store values in matched leaves)
-    abstract fun match(node: IntermediateFormTreeNode): Pair<List<IntermediateFormTreeNode>, Map<String, Any>>?
+    abstract fun match(node: IntermediateFormTreeNode): MatchResult?
 
     // This class is used to match arguments in nodes which aren't other nodes
     // It's used when we want to create instructions for specific values of arguments (constants, labels or registers)
@@ -37,7 +39,7 @@ sealed class IFTPattern {
     }
 
     // Matches any value that fulfill a provided predicate and maps it to the provided name
-    data class ArgumentWhere<T>(val name: String? = null, val predicate: (T) -> Boolean) : ArgumentPattern<T>() {
+    data class ArgumentWhere<T>(val name: String? = null, private val predicate: (T) -> Boolean) : ArgumentPattern<T>() {
         override fun match(value: T): Map<String, Any>? {
             if (!predicate(value)) return null
             return if (name == null) emptyMap() else mapOf(name to value as Any)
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/compiler/intermediate_form/Instruction.kt
```diff
@@ -1,61 +1,152 @@
...
+        }
+
+        // Bitwise instructions
+        data class NotR(override val reg: Register) : OneRegIns() //                                NOT  reg          reg  <- ~reg
+        data class AndRR(override val reg0: Register, override val reg1: Register) : TwoRegIns() // AND  reg0, reg1   reg0 <- reg0 & reg1
+        data class OrRR(override val reg0: Register, override val reg1: Register) : TwoRegIns() //  OR   reg0, reg1   reg0 <- reg0 & reg1
+        data class XorRR(override val reg0: Register, override val reg1: Register) : TwoRegIns() // XOR  reg0, reg1   reg0 <- reg0 ^ reg1
+        data class ShiftLeftR(val reg: Register) : InPlaceInstruction() { //                        SAL  reg,  CL     reg  <- reg << CL
+            override val regsDefined: Collection<Register> = setOf(reg)
+            override val regsUsed: Collection<Register> = setOf(Register.RCX)
```
bebidek, line 115: We **use** reg here as well. Same with right shift.

app/src/main/kotlin/compiler/intermediate_form/Instruction.kt
```diff
@@ -1,61 +1,152 @@
...
+        sealed class SetInstruction : InPlaceInstruction() {
+            abstract val reg: Register
+            override val regsDefined: Collection<Register> get() = setOf(reg)
+            // I'm unsure if this is needed
+            // The SETcc instructions set the lower 8 bits of a register
+            // so the value after SETcc is dependent on the value of reg before instruction.
+            // Specifically, if this wasn't included in when doing
+            // 1: XOR    reg,   reg   ; reset to zero
+            // 2: CMP    rega,  regb
+            // 3: SETcc  reg          ; set lower 8 bits depending on comparison
```
bebidek, line 140: Docs say (and GCC confirms that, so it'll probably be the same with NASM) that operation SETcc will not accept 64-bit register at all.

bebidek (commented):
Few comments regarding the assembly instructions themselves.

JHodorHodor (changes requested):
Kilka stylistycznych komentarzy, plus rzeczy od Bartka. Poza tym wygląda to bardzo dobrze.

app/src/main/kotlin/compiler/intermediate_form/Instruction.kt
rkilar: Yeah, I know that SETcc takes 8-bit registers. I meant for Setcc(reg) to represent SETcc with low 8-bit register corresponding to reg as an operand. I read in documentation that the upper bits are unchanged for 8-bit operands. I'm not sure if the byte order is correct (so setting AL to 1 is the same as setting RAX to 1). I haven't tested it and I'm not all that versed in assembly.

bebidek: According to AMD64 docs, smaller registers (such as AL, AX, EAX) occupy lower bits of full register, so the byte order should be correct.

app/src/main/kotlin/compiler/intermediate_form/IFTPattern.kt
```diff
@@ -0,0 +1,174 @@
+package compiler.intermediate_form
+
+import kotlin.reflect.KClass
+import kotlin.reflect.safeCast
+
+// Class that represents a pattern of an intermediate form tree
+// Not to be confused with the Pattern interface
+sealed class IFTPattern {
+
+    data class MatchResult(val subtrees: List<IntermediateFormTreeNode>, val args: Map<String, Any>)
```
JHodorHodor, line 10:
```suggestion
    data class MatchResult(val unmatchedSubtrees: List<IntermediateFormTreeNode>, val matchedValues: Map<String, Any>)
```
maybe?

JHodorHodor (approved):
Lgtm, dodałem jeszcze jeden malutki komentarz. Po rozwiązaniu kwesti z komentarza Bartka można mergować.

PR #118, opened 2022-12-17
Constant placeholders

tomekb234:


app/src/main/kotlin/compiler/common/Constant.kt
```diff
@@ -0,0 +1,17 @@
+package compiler.common
+
+interface Constant {
```
woocash2, line 3: I think we could instead have `UnspecifiedConstant` as `IFTNode` but this also looks reasonable.

woocash2: But then we'd probably need to have `Add(Const(variables offset), UnspConst(spilled offset))` in `prologue` and `epilogue` and instead we only have `Const` which is more efficient. So that's perfectly fine.

woocash2 (approved):
Looking good. Just make sure that everywhere we used `Long` we now use `Constant`.

PR #119, opened 2022-12-18
Handle spilled registers

JHodorHodor:


app/src/main/kotlin/compiler/intermediate_form/SpillsHandlingAllocation.kt
```diff
@@ -1,20 +1,150 @@
 package compiler.intermediate_form
 
+import compiler.common.intermediate_form.Allocation
+
 object SpillsHandlingAllocation {
 
-    data class AllocationResultWithSpilledOffset(
+    data class Result(
```
woocash2, line 7: Good idea, this long name is not needed

app/src/main/kotlin/compiler/intermediate_form/SpillsHandlingAllocation.kt
```diff
@@ -1,20 +1,150 @@
...
+
+            reservedRegistersNumber ++
+        }
+    }
+
+    data class SpilledInstruction(
+        val instruction: Instruction,
+        val spilledUsedRegisters: Set<Register>,
+        val spilledDefinedRegisters: Set<Register>,
+        val numberOfSpilledRegisters: Int = (spilledUsedRegisters + spilledDefinedRegisters).size,
```
woocash2, line 67: Does this `+` work like a union of sets?

JHodorHodor: yes

app/src/main/kotlin/compiler/intermediate_form/SpillsHandlingAllocation.kt
```diff
@@ -1,20 +1,150 @@
...
+    ): Map<Register, ULong> {
+        val spillsPossibleColors = spilledRegisters.map { Register() }
+        val spillsColoring: Map<Register, Register> = allocator.allocateRegisters(
+            Liveness.LivenessGraphs(
+                Liveness.inducedSubgraph(livenessGraphs.interferenceGraph, spilledRegisters),
+                Liveness.inducedSubgraph(livenessGraphs.copyGraph, spilledRegisters),
+            ),
+            spillsPossibleColors,
+        ).allocatedRegisters
+        val spillsColors = spillsPossibleColors.filter { it in spillsColoring.values }
```
woocash2, line 93: Isn't it just `spollsColoring.values`?

app/src/main/kotlin/compiler/intermediate_form/SpillsHandlingAllocation.kt
```diff
@@ -1,20 +1,150 @@
...
+
+    private fun transformProgram(
+        linearProgram: List<Asmable>,
+        instructionsToSpilled: Map<Int, SpilledInstruction>,
+        allocatedRegisters: Map<Register, Register>,
+        spillsColoring: Map<Register, ULong>,
+    ): List<Asmable> {
+        val newLinearProgram = mutableListOf<Asmable>()
+
+        fun handleSpilledInstruction(spilledInstruction: SpilledInstruction) {
```
woocash2, line 118: Is moving in this function for sure correct?
Now we save each used register to memory before usage and override register with memory after definition.
Shouldn't we save to memory before definition and load from memory before usage

woocash2 (approved):
Looks good, accepted but take a look at one comment on `handleSpilledInstruction` to make sure it's correct.

app/src/main/kotlin/compiler/intermediate_form/SpillsHandlingAllocation.kt
JHodorHodor: Hmm, first, we load to registers to be used from memory (MoveRM), then we perform the instruction and finally, we save defined registers to memory (MoverMR). How can we save to memory before definition (when the thing is not defined yet)?

woocash2: Oh, I see now. `RM` means memory -> register and `MR` register -> memory. I thought it was the other way around.

JHodorHodor: I hope I'm right
```
MoveRM(val reg_dest: Register, val mem_src: Addressing)
MoveMR(val mem_dest: Addressing, val reg_src: Register)
```

woocash2: Yeah, all good.

PR #123, opened 2022-12-21
Refactor the project structure

tomekb234:


app/src/main/kotlin/compiler/lowlevel/allocation/SpillsHandlingAllocation.kt
```diff
@@ -1,6 +1,12 @@
...
+
+import compiler.intermediate.Register
+import compiler.intermediate.generators.memoryUnitSize
+import compiler.lowlevel.Addressing
+import compiler.lowlevel.Asmable
+import compiler.lowlevel.Instruction
+import compiler.lowlevel.Label
+import compiler.lowlevel.dataflow.Liveness
 
 object SpillsHandlingAllocation {
```
JHodorHodor, line 11: Now, it looks a little weird that `SpillsHandlingAllocation` is not an instance of `Allocation`. Maybe we should add an additional interface for this and move all the coloring staff into a nested pachage `partialallocation`. (And also change `Allocation` to `PartialAllocation`) And maybe also change `SpillsHandlingAllocation` to `Allocation` or `FullAllocation`

tomekb234: Renamed `Allocation` into `PartialAllocation` and `SpillsHandlingAllocation` into `Allocation`.
Unrelatedly, also renamed `LexerRegexParser` into `TokenRegexParser`.

There is *a lot* more inconsistencies in this project to fix, and this PR only deals with file names.

JHodorHodor (approved):
lgtm

PR #126, opened 2022-12-27
125 implement foreign function interface

bebidek:


bebidek:
At the moment the base branch is #124, this will be changed later.

app/src/main/kotlin/compiler/analysis/BuiltinFunctions.kt
```diff
@@ -1,96 +1,32 @@
...
-            when (global) {
-                is Program.Global.VariableDefinition -> { analyzeVariable(global.variable) }
-                is Program.Global.FunctionDefinition -> { analyzeFunction(global.function, ::analyzeStatement) }
-            }
-        }
-
-        program.globals.forEach { analyzeGlobal(it) }
-        return result
+        }.toSet()
+        return Program(builtinFunctions.filter { it.key !in globalNames }.map { Program.Global.FunctionDefinition(it.value) } + program.globals)
```
tomekb234, line 30: Why allow overriding built-in functions?

app/src/main/kotlin/compiler/syntax/LanguageTokens.kt
```diff
@@ -30,6 +30,8 @@ object LanguageTokens {
             Pair(TokenType.RETURN, "zwróć"),
             Pair(TokenType.RETURN_UNIT, "zakończ"),
             Pair(TokenType.FUNCTION, "czynność"),
+            Pair(TokenType.FOREIGN, "obca"),
```
tomekb234, line 33: Ja bym nazwał `zewnętrzna` (od `extern`).

app/src/main/kotlin/compiler/syntax/AstFactory.kt
```diff
@@ -50,6 +52,25 @@ object AstFactory {
         return (properNode.children.first() as ParseTree.Leaf).content
     }
 
+    private fun extractForeignName(parseTree: ParseTree.Leaf<Symbol>, diagnostics: Diagnostics, asIdentifier: Boolean): String {
+        return when (parseTree.token()) {
+            TokenType.IDENTIFIER -> parseTree.content
+            TokenType.FOREIGN_NAME -> {
+                val identifier = parseTree.content.substring(1, parseTree.content.lastIndex)
+                if (asIdentifier) {
+                    val walk = RegexDfa(TokenRegexParser.parseStringToRegex("""\l[\l\u\d_]*""")).newWalk()
```
tomekb234, line 61: Would be more efficient to only call this `parseStringToRegex` once.

bebidek: Right.

app/src/main/kotlin/compiler/syntax/LanguageTokens.kt
bebidek: I was considering this name too. I choose "obca" because that one would make one of our example programs invalid.

bebidek: But I agree that "zewnętrzna" sounds better, so I'll change the example program.

app/src/main/kotlin/compiler/analysis/BuiltinFunctions.kt
bebidek: Do we assume that?
I treat them as functions declared at level -1, which means that global functions (at level 0) can override them.

tomekb234: If we were to add some module system, then it would make sense for the standard library to be some implicitly imported module, but the import statement is usually understood to introduce symbols into the global scope. The existence of "level -1" is a bit at odds with the meaning of the word "global" in "global scope".

tomekb234: (In the sense that the global scope is usually *the uppermost* scope.)

tomekb234: But I suppose this consistency is not that much important - we can make an exception for the "implicitly imported" symbols just for the sake of convenience. You can leave this is as it is if the arguments above are not convincing.

tomekb234 (approved):

app/src/main/kotlin/compiler/analysis/BuiltinFunctions.kt
bebidek: Well, I treated "standard library" and "builtin functions" as two different things: "Standard library" could potentially have more functions which are not imported by default and "Builtin functions" as subset of standard library and these function are imported by the compiler (potentially in different way than "global import" you mentioned).
But if it significantly changes our plans then I don't insist.

tomekb234: We don't really have any plans about these matters yet, only assumptions. And this distinction between "standard library" and "built-in functions" seems quite reasonable, so let's leave this as it is.

PR #127, opened 2022-12-27
Implement toAsm

JHodorHodor:


app/src/main/kotlin/compiler/intermediate/Register.kt
```diff
@@ -1,23 +1,25 @@
 package compiler.intermediate
 
 // A class representing virtual cpu register
-class Register {
+class Register(private val asmName: String = "VIRTUAL REGISTER") {
```
tomekb234, line 4: I think we shouldn't have placeholder name for virtual registers, but just throw an error when one is attempted to be printed.

app/src/main/kotlin/compiler/intermediate/Register.kt
```diff
@@ -1,23 +1,25 @@
...
+    override fun toString() = asmName
```
tomekb234, line 5: It would be more readable to use a dedicated function like `toAsm` instead of the generic `toString`.

app/src/main/kotlin/compiler/lowlevel/Instruction.kt
```diff
@@ -19,12 +19,25 @@ sealed class Instruction : Asmable {
...
+        override fun toAsm(registers: Map<Register, Register>) = when (this) {
+            is JmpEq -> "je"
+            is JmpNEq -> "jne"
+            is JmpLt -> "jl"
+            is JmpLtEq -> "jle"
+            is JmpGt -> "jg"
+            is JmpGtEq -> "jge"
+            is JmpZ -> "jz"
+            is JmpNZ -> "jnz"
+            is Dummy -> "dummy-conditional-jump"
```
tomekb234, line 38: We should throw an exception here.

app/src/main/kotlin/compiler/lowlevel/Label.kt
```diff
@@ -4,7 +4,5 @@ import compiler.intermediate.Register
 import java.io.PrintWriter
 
 data class Label(val label: String) : Asmable {
-    override fun writeAsm(output: PrintWriter, registers: Map<Register, Register>) {
-        TODO()
-    }
+    override fun writeAsm(output: PrintWriter, registers: Map<Register, Register>) = output.write("$label:")
```
tomekb234, line 7: These labels should be local (prefixed with `.`).

app/src/main/kotlin/compiler/lowlevel/Instruction.kt
```diff
@@ -152,13 +227,25 @@ sealed class Instruction : Asmable {
...
             override val regsUsed: Collection<Register> get() = setOf(reg)
+
+            override fun toAsm(registers: Map<Register, Register>) = when (this) {
+                is SetEqR -> "sete"
+                is SetNeqR -> "setne"
+                is SetLtR -> "setl"
+                is SetLtEqR -> "setle"
+                is SetGtR -> "setg"
+                is SetGtEqR -> "setge"
+            } + " ${registers[reg]}"
```
bebidek, line 238: These instructions require 8-bit register (e.g. "sete al", "sete rax" would fail), so some translation will be necessary.

app/src/main/kotlin/compiler/intermediate/Register.kt
JHodorHodor: You're right, changed.

app/src/main/kotlin/compiler/intermediate/Register.kt
JHodorHodor: Done

app/src/main/kotlin/compiler/lowlevel/Instruction.kt
JHodorHodor: True, I added a mapping.

tomekb234 (approved):

PR #128, opened 2022-12-27
Add explicit register usage info to Call/Ret + bugfix

bebidek:
This is my suggestion how to solve an issue we discussed on Discord. Also, I remove some code from Linearization that is no longer necessary (actually, it wasn't necessary even before these changes, but now it's more clear why).

### DIFF
--- a/app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
+++ b/app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
@@ -172,6 +172,12 @@ data class DefaultFunctionDetailsGenerator(
             IFTNode.RegisterWrite(Register.RBP, IFTNode.StackPop())
         )
 
+        // return
+        cfgBuilder.addLinksFromAllFinalRoots(
+            CFGLinkType.UNCONDITIONAL,
+            IFTNode.Return(setOf(Register.RAX) + calleeSavedRegistersWithoutRSPAndRBP)
+        )
+
         return cfgBuilder.build()
     }
 
--- a/app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
+++ b/app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
@@ -113,9 +113,11 @@ data class DefaultFunctionDetailsGenerator(
                 CFGLinkType.UNCONDITIONAL,
                 genWrite(
                     param.value,
-                    IFTNode.Add(
-                        IFTNode.RegisterRead(Register.RBP), // add 2 to account old RBP and return address
-                        IFTNode.Const((param.index + 2) * memoryUnitSize.toLong())
+                    IFTNode.MemoryRead(
+                        IFTNode.Add(
+                            IFTNode.RegisterRead(Register.RBP), // add 2 to account old RBP and return address
+                            IFTNode.Const((param.index + 2) * memoryUnitSize.toLong())
+                        )
                     ),
                     true
                 ),
--- a/app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
+++ b/app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
@@ -126,10 +126,10 @@ data class DefaultFunctionDetailsGenerator(
         }
 
         // backup callee-saved registers
-        for ((hardware_reg, backup_reg) in calleeSavedRegistersWithoutRSPAndRBP zip calleeSavedBackupRegisters)
+        for ((hardwareReg, backupReg) in calleeSavedRegistersWithoutRSPAndRBP zip calleeSavedBackupRegisters)
             cfgBuilder.addLinksFromAllFinalRoots(
                 CFGLinkType.UNCONDITIONAL,
-                IFTNode.RegisterWrite(backup_reg, IFTNode.RegisterRead(hardware_reg))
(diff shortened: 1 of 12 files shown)


app/src/main/kotlin/compiler/intermediate/generators/SysV64CallingConvention.kt
```diff
@@ -27,16 +30,19 @@ object SysV64CallingConvention {
         // Add call instruction to actually call a given function
         cfgBuilder.addLinksFromAllFinalRoots(
             CFGLinkType.UNCONDITIONAL,
-            IFTNode.Call(targetFunction)
+            IFTNode.Call(targetFunction, usedRegisters, callerSavedRegisters)
```
tomekb234, line 33: All these various words "used", "defined" and "saved" used interchangeably are somewhat confusing. Maybe just use "read" and "written"?

app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
```diff
@@ -35,6 +35,7 @@ data class DefaultFunctionDetailsGenerator(
     private val variablesRegisters: MutableMap<NamedNode, Register> = referenceHashMapOf()
     private var variablesTotalOffset: ULong = 0u
     private val previousDisplayEntryRegister = Register()
+    private val calleeSavedBackupRegisters = calleeSavedRegistersWithoutRSPAndRBP.map { Register() }
```
tomekb234, line 38: Since we are at it, why not backup RBP in this way as well?

app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
```diff
@@ -113,20 +114,22 @@ data class DefaultFunctionDetailsGenerator(
...
+                        )
                     ),
                     true
                 ),
             )
         }
 
         // backup callee-saved registers
-        for (register in calleeSavedRegistersWithoutRSPAndRBP.reversed())
+        for ((hardware_reg, backup_reg) in calleeSavedRegistersWithoutRSPAndRBP zip calleeSavedBackupRegisters)
```
tomekb234, line 129: :snake: → :camel:

app/src/main/kotlin/compiler/intermediate/generators/SysV64CallingConvention.kt
bebidek: Is it that confusing? I would say it's less confusing this way as these words have slightly different meanings. The terms "caller-saved" and "callee-saved" are part of calling convention docs (I consider them well-known) and they are more high-level terms (they describe function as a whole). On the other side, the meaning of "used" and "defined" is taken from liveness problem, and we use them to describe single instructions.
Here, variable `usedRegisters` hold list of registers which are going to be **used** by Call instruction.

app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
bebidek: It took me some time (plus help from my brother) to understand this comment, but of course you're right.

app/src/main/kotlin/compiler/intermediate/generators/SysV64CallingConvention.kt
tomekb234: Yeah, makes sense.

app/src/main/kotlin/compiler/lowlevel/linearization/Linearization.kt
```diff
@@ -77,34 +76,14 @@ object Linearization {
                     dfs(targetWhenTrue, label)
                     dfs(targetWhenFalse, nextLabel)
                 }
-            } else if (node in cfg.conditionalTrueLinks) {
```
tomekb234, line 80: I mentioned about this here: https://github.com/TCS-Compilers/compiler/pull/114#discussion_r1045985093, but I guess we can remove this for now.

app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
bebidek: Old RBP must be pushed to the stack right below return address (according to the calling convention).

app/src/test/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGeneratorTest.kt
```diff
@@ -20,6 +20,9 @@ class DefaultFunctionDetailsGeneratorTest {
     private val resultDummyVariable = Variable(Variable.Kind.VALUE, "numberResult", Type.Number, null)
     private val resultVariableRegister = Register()
 
+    private val callerSavedRegisters = listOf(Register.RAX, Register.RCX, Register.RDX, Register.RSI, Register.RDI, Register.R8, Register.R9, Register.R10, Register.R11)
```
tomekb234, line 23: Why copy the list from `SysV64CallingConvention`?

tomekb234 (approved):

app/src/test/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGeneratorTest.kt
bebidek: The list in `SysV64CallingConvention` is part of the very implementation we want to test. Let's imagine that someone mistakenly changed the list in SysV64CallingConvention.kt. In that case, some unit tests should fail but if we imported the list in test files, we wouldn't detect that.

PR #129, opened 2022-12-27
Add simple VIM syntax file

bebidek:
As we haven't decided on the name of our language, the temporary one is "nienazwany".


tomekb234 (changes requested):
I would rather name these files as `extras/vim/ftdetect/[language name].vim` and `extras/vim/syntax/[language name].vim` (for easier copying). We can keep the temporary name "nienazwany" (short form `nie`) for now.

tomekb234 (approved):

PR #130, opened 2022-12-29
Display asmable

woocash2:


app/src/main/kotlin/compiler/lowlevel/storage/DisplayStorage.kt
```diff
@@ -5,6 +5,9 @@ import java.io.PrintWriter
 
 class DisplayStorage(val program: Program) {
     fun writeAsm(output: PrintWriter) {
-        TODO() // Reserve memory for the display with appropriate maximum depth
+        output.write("extern malloc")
```
tomekb234, line 8: This is supposed to be in the `.bss` section, at the label used by `DefaultFunctionDetailsGenerator`.

app/src/main/kotlin/compiler/ast/Program.kt
```diff
@@ -3,6 +3,8 @@ package compiler.ast
 import compiler.input.LocationRange
 
 data class Program(val globals: List<Global>) {
+    var staticFunctionDepth: Int = 0
```
tomekb234, line 6: This would be the only mutable variable in AST, which would be quite odd. The maximum depth is only used by `DisplayStorage`, so it may as well be calculated there. Alternatively, `DisplayStorage` may receive additional constructor argument with the maximum depth calculated somewhere else.

woocash2: Writing another ast traversal would be redundant. I'm gonna make name resolution return depth as part of the result and include it in `ProgramProperties`. We can then pass depth to `DisplayStorage` constructor.

app/src/main/kotlin/compiler/lowlevel/storage/DisplayStorage.kt
woocash2: How should we call this label?

app/src/main/kotlin/compiler/lowlevel/storage/DisplayStorage.kt
```diff
@@ -1,10 +1,12 @@
...
 import java.io.PrintWriter
 
-class DisplayStorage(val program: Program) {
+class DisplayStorage(private val programStaticDepth: Int) {
+
+    val displayLabel: String = "display"
+
     fun writeAsm(output: PrintWriter) {
-        TODO() // Reserve memory for the display with appropriate maximum depth
+        output.write("$displayLabel: resq $programStaticDepth")
```
woocash2, line 10: assumed `section .bss` is added somewhere before, because this can be not the only thing in this section.

app/src/main/kotlin/compiler/lowlevel/storage/DisplayStorage.kt
```diff
@@ -1,10 +1,12 @@
 package compiler.lowlevel.storage
 
-import compiler.ast.Program
 import java.io.PrintWriter
 
-class DisplayStorage(val program: Program) {
+class DisplayStorage(private val programStaticDepth: Int) {
+
+    val displayLabel: String = "display"
```
tomekb234, line 7: This label should not be defined in two places (here and in `DefaultFunctionDetailsGenerator.kt`).

woocash2: Didn't see it. I'm gonna remove it from details generator.

tomekb234: Actually, the other way around wold be preferable, so that `intermediate` packages does not depend on `lowlevel`.

woocash2: ok ok

woocash2: it was more intuitive to me this way

tomekb234: I guess the direction of dependencies between packages in this project (low-level to high-level or the other way around) is debatable, but let's keep it as it is for now.

tomekb234 (approved):

PR #131, opened 2022-12-29
Storage for global variables

kamgal00:


tomekb234 (changes requested):
I just realized that we don't handle `Variable.Kind.CONSTANT` in the way it is supposed to be handled. The constants should be replaced with their values at compiation time and they should not reserve any additional memory. To implement this it should be sufficient to filter out constants in `GlobalVariableAccessGenerator` and `GlobalVariableStorage`, and to add a special case in `GlobalVariableAccessGenerator.genRead` to return IFT nodes with constants instead of memory reads (the function `getValueOfConstExpr` can be reused for this purpose).

app/src/main/kotlin/compiler/lowlevel/storage/GlobalVariableStorage.kt
```diff
@@ -1,10 +1,29 @@
...
+        null -> 0
+        is Expression.UnitLiteral -> 0
+        is Expression.BooleanLiteral -> if (expr.value) 1 else 0
+        is Expression.NumberLiteral -> expr.value
+        else -> throw IllegalArgumentException()
+    }
+
     fun writeAsm(output: PrintWriter) {
-        TODO() // Reserve memory for global variables with appropriate initial values
+        output.write("section .data\n")
```
tomekb234, line 20: I think we can assume that this line will be written by some other code.

tomekb234 (approved):

app/src/main/kotlin/compiler/ast/Expression.kt
```diff
@@ -113,4 +113,13 @@ sealed class Expression : AstNode {
         val resultWhenFalse: Expression,
         override val location: LocationRange? = null,
     ) : Expression()
+
+    companion object {
+        fun getValueOfLiteral(expr: Expression) = when (expr) {
+            is UnitLiteral -> 0L
```
bebidek, line 119: We have a constant for that (in IFTNode.kt).

tomekb234: Actually, unit values should not even exist in the generated code, so whichever representation is chosen does not really matter.

bebidek: I agree with that, it just looks little inconsistent if we have a constant UNIT_VALUE and use it in some places while in the others we have literal `0L`. So maybe it would be better idea to remove the constant (since it has no semantic meaning).

PR #133, opened 2022-12-30
Make sure stack is aligned to 16 bytes

ggawryal:
We're using stack pointer modifying instructions (add directly to `RSP`, `StackPush`, `StackPop` and `Call`) only in `genPrologue, genEpilogue` and `genCall` functions. 
The stack will be alligned to 16 bytes at the entry of each program, so all we have to do is to:
1) move RSP by 8 bytes at thebeginning of `genCall` if some function has more than an odd, larger than 6 number of arguments
2) make sure allocated space for non-local and spilled variables in `genPrologue` is divisible by 16.

### DIFF
--- a/app/src/main/kotlin/compiler/intermediate/Constant.kt
+++ b/app/src/main/kotlin/compiler/intermediate/Constant.kt
@@ -16,9 +16,10 @@ class SummedConstant(private val first: Constant, private val second: Constant) 
     override val value get() = first.value + second.value
 }
 
-class ConstantAlignedToAGivenRestModulo(private val rawValue: Constant, private val modulo: Long, private val rest: Long = 0L) : Constant {
+class AlignedConstant(private val rawValue: Constant, private val modulo: Long, private val remainder: Long = 0L) : Constant {
+    // smallest number greater or equal to rawValue having a given remainder modulo
     override val value: Long get() {
-        val toAdd = rest - rawValue.value % modulo
+        val toAdd = remainder.mod(modulo) - rawValue.value % modulo
         return if (toAdd < 0) rawValue.value + toAdd + modulo
         else rawValue.value + toAdd
     }
--- a/app/src/main/kotlin/compiler/intermediate/Constant.kt
+++ b/app/src/main/kotlin/compiler/intermediate/Constant.kt
@@ -15,3 +15,11 @@ class SummedConstant(private val first: Constant, private val second: Constant) 
 
     override val value get() = first.value + second.value
 }
+
+class ConstantAlignedToAGivenRestModulo(private val rawValue: Constant, private val modulo: Long, private val rest: Long = 0L) : Constant {
+    override val value: Long get() {
+        val toAdd = rest - rawValue.value % modulo
+        return if (toAdd < 0) rawValue.value + toAdd + modulo
+        else rawValue.value + toAdd
+    }
+}
--- a/app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
+++ b/app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
@@ -3,6 +3,7 @@ package compiler.intermediate.generators
 import compiler.ast.NamedNode
 import compiler.ast.Variable
 import compiler.intermediate.CFGLinkType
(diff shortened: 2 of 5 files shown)


app/src/main/kotlin/compiler/intermediate/Constant.kt
```diff
@@ -15,3 +15,11 @@ class SummedConstant(private val first: Constant, private val second: Constant)
 
     override val value get() = first.value + second.value
 }
+
+class ConstantAlignedToAGivenRestModulo(private val rawValue: Constant, private val modulo: Long, private val rest: Long = 0L) : Constant {
```
tomekb234, line 19: I think the word "remainder" is used instead of "rest" in this context.

tomekb234 (approved):

app/src/main/kotlin/compiler/intermediate/Constant.kt
tomekb234: I'd suggest to simply name this as `AlignedConstant`.

app/src/main/kotlin/compiler/intermediate/generators/SysV64CallingConvention.kt
```diff
@@ -7,41 +7,59 @@ import compiler.intermediate.Register
...
+        val cfgBuilder = ControlFlowGraphBuilder()
+        // Align stack
+        if (stackMovement % stackAlignmentInBytes != 0L) {
+            val toAdd = stackAlignmentInBytes - stackMovement % stackAlignmentInBytes
+            cfgBuilder.addLinksFromAllFinalRoots(
+                CFGLinkType.UNCONDITIONAL,
+                IFTNode.Subtract(
+                    IFTNode.RegisterRead(Register.RSP),
+                    IFTNode.Const(toAdd)
+                )
```
bebidek, line 40: You need to actually save the result to RSP (`IFTNode.RegisterWrite`).

app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
```diff
@@ -67,7 +68,7 @@ data class DefaultFunctionDetailsGenerator(
         cfgBuilder.addLinksFromAllFinalRoots(
             CFGLinkType.UNCONDITIONAL,
             IFTNode.StackPush(IFTNode.RegisterRead(Register.RBP))
-        )
+        ) // disaligns stack
```
bebidek, line 71: Actually, this PUSH aligns the stack to 16B. SysV64 convention says that right after CALL, RSP % 16 = 8, so the push makes it aligned.

bebidek: Consequently, I am not sure whether the  whole program is correct as:
At this point (after PUSH) stack is aligned.
After local variable allocation stack is 8 modulo 16.
After pushing the arguments (while calling some other function) the stack is still 8 modulo 16.
After CALL, stack is aligned.
After PUSH, stack is not aligned.

bebidek (changes requested):

app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
tomekb234: Indeed, it seems that right now CALLs are at 8 mod 16. Changing the remainder to 0 at line 90 should fix the problem.

app/src/main/kotlin/compiler/intermediate/Constant.kt
ggawryal: You're right. Applied both suggestions and added a short doc comment

app/src/main/kotlin/compiler/intermediate/generators/SysV64CallingConvention.kt
ggawryal: Great catch, fixed

app/src/main/kotlin/compiler/intermediate/generators/DefaultFunctionDetailsGenerator.kt
ggawryal: Thanks for spotting, fixed

bebidek (approved):

PR #134, opened 2022-12-30
Compiler integration

rkilar:
There  are some areas that need work. After changing `ControlFlowGraph.treeRoots` to use `ReferenceHashMap` instead of `List`, current test don't throw errors. However, the code that's generated has some problems:
- The register allocation sometimes uses the `rsp` register which causes segfaults. I don't think there's an option to  ban certain registers from being used in the current code or give a partial coloring in the current code.
- There are probably some bugs in liveness/allocation. For example for the program:
`czynność f() -> Liczba {
    jeśli(prawda)
        zwróć 12
    wpp zwróć 12
}
`
In one of the cases the the instruction `mov rax, 12` is removed from the program despite being generated by linearization. This doesn't happen when the return values are different suggesting some code might be comparing the instructions by value.

I'm also not really sure about how to link the standard library. Currently the location of stdlib.c has to be passed as an argument same as other external code. This is not optimal, but I'm not sure what's the best way to do it.

### DIFF
--- a/app/src/main/kotlin/compiler/App.kt
+++ b/app/src/main/kotlin/compiler/App.kt
@@ -1,6 +1,24 @@
 package compiler
 
+import compiler.diagnostics.CompilerDiagnostics
+import java.io.PrintWriter
+import java.io.StringReader
+import java.io.StringWriter
+
 fun main() {
+    val diagnostics = CompilerDiagnostics()
+    val compiler = Compiler(diagnostics)
+
+    val program = """
+        czynność główna() {
+            zm n : Liczba = 3
+            n = n + 1
+        }
+    """
+    val reader = StringReader(program)
+    val writer = PrintWriter(StringWriter())
+
+    compiler.process(reader, writer)
     // TODO:
     // - create Reader for input source file
     // - create Writer for output assembly file
--- a/app/src/main/kotlin/compiler/App.kt
+++ b/app/src/main/kotlin/compiler/App.kt
@@ -7,49 +7,48 @@ import com.github.ajalt.clikt.parameters.options.option
 import compiler.diagnostics.CompilerDiagnostics
 import java.io.File
 import java.io.FileNotFoundException
-import java.io.FileReader
-import java.io.PrintWriter
+import kotlin.system.exitProcess
 
 fun runCommand(command: String) {
     println(command)
-    ProcessBuilder(*command.split(" ").toTypedArray())
+    val process = ProcessBuilder(*command.split(" ").toTypedArray())
         .redirectOutput(ProcessBuilder.Redirect.INHERIT)
         .redirectError(ProcessBuilder.Redirect.INHERIT)
         .start()
+    process.waitFor()
+    if (process.exitValue() != 0) {
+        exitProcess(process.exitValue())
+    }
 }
 
 class Cli : CliktCommand() {
-    val inputFile: String by argument("input file", help = "Input file")
-    val externFile: String? by option("-e", "--extern", help = "Extern file")
-    val assemblyFile: String by option("-a", "--assembly", help = "Assembly file").default("a.asm")
-    val listingFile: String? by option("-l", "--listing", help = "Listing file")
-    val objectFile: String by option("-b", "--object", help = "Object file").default("a.o")
-    val outputFile: String by option("-o", "--output", help = "Output file").default("a.out")
+    private val inputFile: String by argument("input file", help = "Input file")
+    private val stdlibLocation: String by option("-l", "--stdlib", help = "Standard library location").default("stdlib.o")
+    private val assemblyFile: String by option("-a", "--assembly", help = "Assembly file").default("a.asm")
+    private val objectFile: String by option("-b", "--object", help = "Object file").default("a.o")
+    private val outputFile: String by option("-o", "--output", help = "Output file").default("a.out")
 
     override fun run() {
         val diagnostics = CompilerDiagnostics()
         val compiler = Compiler(diagnostics)
 
         try {
-            val reader = FileReader(File(inputFile))
-
-            val writer = PrintWriter(File(assemblyFile))
-
-            if (!compiler.process(reader, PrintWriter(writer))) {
-                println("Compilation failed")
-                diagnostics.diagnostics.forEach { println(it) }
-                return
+            File(inputFile).reader().use { input ->
+                File(assemblyFile).writer().use { output ->
+                    if (!compiler.process(input, output)) {
+                        println("Compilation failed")
+                        diagnostics.diagnostics.forEach { println(it) }
+                        exitProcess(1)
+                    }
+                }
             }
-
-            writer.close()
-            reader.close()
         } catch (err: FileNotFoundException) {
-            println("Error: ${err.message}")
-            return
+            println("File error: ${err.message}")
+            exitProcess(1)
         }
 
-        runCommand("nasm -f elf64 $assemblyFile${listingFile?.let { " -l $it" } ?: ""} -o $objectFile")
-        runCommand("gcc -static $objectFile${externFile?.let {" $it"} ?: ""} -o $outputFile")
+        runCommand("nasm -f elf64 $assemblyFile -o $objectFile")
+        runCommand("gcc -static $objectFile $stdlibLocation -o $outputFile")
     }
 }
 
--- a/app/src/main/kotlin/compiler/App.kt
+++ b/app/src/main/kotlin/compiler/App.kt
@@ -1,29 +1,56 @@
 package compiler
 
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.arguments.argument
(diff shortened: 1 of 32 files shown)


tomekb234:
The standard library should be in some fixed system-wide location, such as `/usr/lib`, ideally installed by some package manager. Alternatively, it can be added as a resource (at `app/src/main/resources` or so), and then it will be available at runtime (even if the program is packaged as a portable JAR), and it can be copied to some temporary location, like `/tmp` or `~/.cache`.

For the time being, I think it should be sufficient to just assume that there is `stdlib.o` in the current directory.

app/src/main/kotlin/compiler/App.kt
```diff
@@ -1,11 +1,56 @@
...
+    println(command)
+    ProcessBuilder(*command.split(" ").toTypedArray())
+        .redirectOutput(ProcessBuilder.Redirect.INHERIT)
+        .redirectError(ProcessBuilder.Redirect.INHERIT)
+        .start()
+}
+
+class Cli : CliktCommand() {
+    val inputFile: String by argument("input file", help = "Input file")
+    val externFile: String? by option("-e", "--extern", help = "Extern file")
```
tomekb234, line 23: I would name this argument as `stdlib` and (temporarily) assume the default value of `stdlib.o` in the current directory.

app/src/main/kotlin/compiler/App.kt
```diff
@@ -1,11 +1,56 @@
...
+    override fun run() {
+        val diagnostics = CompilerDiagnostics()
+        val compiler = Compiler(diagnostics)
+
+        try {
+            val reader = FileReader(File(inputFile))
+
+            val writer = PrintWriter(File(assemblyFile))
+
+            if (!compiler.process(reader, PrintWriter(writer))) {
```
tomekb234, line 38: Unnecessary wrapping with `PrintWriter`.

app/src/main/kotlin/compiler/App.kt
```diff
@@ -1,11 +1,56 @@
...
+                println("Compilation failed")
+                diagnostics.diagnostics.forEach { println(it) }
+                return
+            }
+
+            writer.close()
```
tomekb234, line 44: It would be safer to use `reader.use` and `writer.use`.

app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -24,8 +37,9 @@ class Compiler(val diagnostics: Diagnostics) {
 
     private val lexer = Lexer(LanguageTokens.getTokens(), diagnostics)
     private val parser = Parser(LanguageGrammar.getGrammar(), diagnostics)
+    private val covering = DynamicCoveringBuilder(InstructionSet.getInstructionSet())
 
-    fun process(input: Reader, output: Writer): Boolean {
+    fun process(input: Reader, output: PrintWriter): Boolean {
```
tomekb234, line 42: It could have remained as `Writer` (for symmetry with `Reader`), and then it could be wrapped into `PrintWriter`.

app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -40,17 +54,72 @@ class Compiler(val diagnostics: Diagnostics) {
...
+                astWithBuiltinFunctions,
+                programProperties.variableProperties,
+                programProperties.functionReturnedValueVariables,
+                diagnostics.hasAnyError()
+            )
+
+            val functionCFGs = createGraphForProgram(astWithBuiltinFunctions, programProperties, functionDetailsGenerators, diagnostics)
+
+            val mainFunction = (
+                ast.globals.find { it is Program.Global.FunctionDefinition && it.function.name == "główna" } as Program.Global.FunctionDefinition?
```
tomekb234, line 67: The name of the main function should be a global constant, for better readability.

app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -40,17 +54,72 @@ class Compiler(val diagnostics: Diagnostics) {
...
+                    it.value,
+                    Liveness.computeLiveness(it.value),
+                    Allocation.REGISTER_ORDER,
+                    ColoringAllocation
+                )
+            }
+
+            finalCode.entries.forEach { functionDetailsGenerators[it.key]!!.spilledRegistersOffset.settledValue = it.value.spilledOffset.toLong() }
+
+            output.println("SECTION .bss")
```
tomekb234, line 89: I would suggest to move the code with the actual printing of section headers and the `main` label into to the `lowlevel` package. It could be a function (say, `printAsmFile`), which would take three functions as arguments, corresponding to the three sections, and call these functions after printing the headers of the corresponding sections. In this way we would have a readable template with the layout of output assembly file.

app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -40,17 +54,72 @@ class Compiler(val diagnostics: Diagnostics) {
...
+            output.println("\nSECTION .text")
+            functionDetailsGenerators
+                .filter { it.key.implementation is Function.Implementation.Foreign }
+                .forEach { output.println("extern ${it.value.identifier}") }
+            output.println(
+                """
+                    global main
+                    main:
+                        call ${functionDetailsGenerators[mainFunction]!!.identifier}
+                        ret
```
bebidek, line 104: Doing this that way will break the stack alignment. We should either use JMP here or construct valid stack frame (i.e. "push rbp" followed by "mov rbp, rsp").

app/src/main/kotlin/compiler/lowlevel/allocation/Allocation.kt
```diff
@@ -10,6 +10,25 @@ import compiler.lowlevel.dataflow.Liveness
...
+        Register.RSP,
+        Register.R9,
+        Register.R8,
+        Register.RDI,
+        Register.RSI,
+        Register.RDX,
+        Register.RCX,
+        Register.RBX,
+        Register.RAX,
+    )
```
bebidek, line 30: Regarding your concerns about RSP being used in allocation, wouldn't removing it from here solve the problem?

bebidek: Speaking of which, I think RBP should be also excluded from here (at least for now). While it is not that important as RSP (programs should still work correctly), the debugger (which might be useful for us) will have problems (for example with showing a stack trace).

bebidek (commented):
Few more comments.

app/src/main/kotlin/compiler/lowlevel/allocation/Allocation.kt
bebidek: Also, I think the order of these registers is not optimal. The comment in Allocation.kt:38 says "Registers should be sorted by likeliness to be reserved for spilled". R15, R14, R13 are all callee-saved registers, so we should use them for spills last, because we read spilled values from memory every time, so conserving value of spill-handling registers across function calls is pointless, while it's useful for local variables.
I would suggest moving R11, R10 to the top of that list, as these registers are caller-saved and they are not used in parameter passing (which could potentially cause some problems).

bebidek:
One more thing that seems missing is returning non-zero exit code when compilation fails.

app/src/main/kotlin/compiler/lowlevel/allocation/Allocation.kt
bebidek: > Regarding your concerns about RSP being used in allocation, wouldn't removing it from here solve the problem?

OK, I see that it won't fix it, because RSP is being then colored by some other register.

app/src/main/kotlin/compiler/Compiler.kt
tomekb234: We should also return 0 as exit code. (`mov rax, 0`).

bebidek: We could check here whether `główna` function return type is Unit or Number. If it's Number, then we don't touch RAX and treat the return value as exit code. Otherwise, we should indeed set RAX to zero (why not use _idiomatic_ `xor rax, rax` here).

bebidek:
> 

These two problems you listed will be fixed by PR #136, which will be merged after this one.

tomekb234 (approved):

app/src/main/kotlin/compiler/lowlevel/CodeSection.kt
```diff
@@ -18,10 +18,11 @@ class CodeSection(
             """
                 global main
                 main:
-                    mov rbp, rsp
                     push rbp
+                    mov rbp, rsp
                     call $mainFunctionLabel
                     ${if (ignoreMainReturnValue) "xor rax, rax" else "" }
+                    mov rsp, rbp
```
bebidek, line 25: That line doesn't do anything.

rkilar:
Integration fails on one of the tests but I'm not able to replicate it on any of my machines.

PR #135, opened 2022-12-31
Add integration tests

ignacy123:
Integration tests are ignored because they all fail with segfaults so far.


test_all.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+./gradlew test "$@"
```
tomekb234, line 2: A bit odd that the arguments are passed particularly to this one command. I would just hardcode `--no-daemon` here.

run_integration_tests.sh
```diff
@@ -0,0 +1,83 @@
...
+FAILED_TO_COMPILE=0
+
+# bash trickery to get rid of identations at the start of lines
+print_multiline_message() {
+    echo "$1" | sed 's/^[ \t]*//g'
+}
+
+run_test_case() {
+    test_number=$(echo $1 | grep -Eo '[0-9]+$')
+    echo "Case number $test_number..."
```
tomekb234, line 16: I think it's too verbose to print a new line for every successful test. To indicate progress we can just output the number of tests passed with `\r`.

run_integration_tests.sh
```diff
@@ -0,0 +1,83 @@
...
+
+run_test_cases() {
+    csplit --suppress-matched -f "$TEMP_TEST_CASE_PREFIX" -z $1 "/^-*$/" '{*}' 1>/dev/null
+    for test_case in $(find . -name "$TEMP_TEST_CASE_PREFIX*"); do
+        run_test_case $test_case
+        rm $test_case
+    done
+}
+
+./gradlew install
```
tomekb234, line 52: Why is this necessary?

run_integration_tests.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+IGNORE_HEADER="// INTEGRATION TESTS IGNORE"
```
tomekb234, line 2: Why not just `// IGNORED`?

tomekb234 (commented):
Some naming suggestions:

- The directory name `integration_tests` feels unnecessarily long, the name `tests` would be just fine, in my opinion. Since this is outside of `src` directory, it can be guessed from the context that these must be final integration tests.
- It's odd to categorize basic tests as `no_io`, just to show contrast with `io`. If we wanted to add another category of tests, what name should it have? I would just name that directory as `basic`.

test_all.sh
ignacy123: We don't want `--no-daemon` when running locally. This script is just a simple wrapper, I don't think such argument passing is a big issue

run_integration_tests.sh
ignacy123: I wanted to choose something specific enough that for sure will not appear in actual programs... I think it's possible (though highly unlikely) that someone might use `// IGNORED` as an actual comment

.circleci/config.yml
```diff
@@ -10,6 +10,9 @@ jobs:
       - image: cimg/openjdk:11.0
     steps:
       - checkout
+      - run:
+          name: Install dependencies
+          command: sudo apt update && xargs -a app/dependencies.txt sudo apt install
```
tomekb234, line 15: Let's just assume that NASM is already installed (and write in README about this). Also, not everyone uses `apt`...

tomekb234 (approved):

PR #138, opened 2023-01-02
Tidy up the mess with references

tomekb234:
This PR removes `ReferenceMap` (and `ReferenceSet`) and introduces `Ref`. It is simply a wrapper for any object, which overrides `equals` and `hashCode` to compare by identity.

`ReferenceSet<E>` ~(now called `RefSet<E>`) is simply an alias for~ can now be realized with `Set<Ref<E>>`. To access it, explicit wrapping and unwrapping of `Ref`s is necessary, which is less convenient, but should lead to less confusion caused by the violation of `Set`'s invariants. We also no longer need any dedicated utility methods for `ReferenceSet` (such as `combineReferenceSets`), as we can just use Kotlin's standard library. In particular, the comparisons in tests involving `Ref`s should now be quite easier.

The class `Ref` can also serve as a general marker for objects which are `data class`, but in the given context should be distinguished by identity (in particular `AstNode`s, `IFTNode`s and `Instruction`s). For example, in `ControlFlowGraph.unconditionalLinks` *both* keys and values are supposed to be distinguished by identity (in constrast to for example the type checker's result), and so both the key and the value type should be marked with `Ref`, which would also make comparing these maps more convenient. ~Therefore, two type aliases for maps are proposed: `RefMap<K, V>` for `Map<Ref<K>, Ref<V>>`, and `KeyRefMap<K, V>` for `Map<Ref<K>, V>`.~

With this marker, the operator `===` (which is easily confused with `==`) should not be necessary (outside of `Ref`'s implementation). This PR has done some initial refactoring to decrease the seeming confusion caused by distinguishing `data class` objects both by value and by identity (and also refactored a few other minor unrelated things), but still more work is necessary to make the code generally more readable.


app/src/main/kotlin/compiler/utils/Ref.kt
```diff
@@ -0,0 +1,56 @@
...
+typealias RefMap<K, V> = Map<Ref<K>, Ref<V>>
+
+typealias KeyRefMap<K, V> = Map<Ref<K>, V>
+
+typealias MutableRefSet<E> = MutableSet<Ref<E>>
+
+typealias MutableRefMap<K, V> = MutableMap<Ref<K>, Ref<V>>
+
+typealias MutableKeyRefMap<K, V> = MutableMap<Ref<K>, V>
+
```
ignacy123, line 22: Do we need these at all? I think the explicit types are less confusing

app/src/main/kotlin/compiler/ast/AstNode.kt
```diff
@@ -28,6 +28,8 @@ sealed interface AstNode {
                 is Function -> "czynność ${this.name}(${this.parameters.joinToString { it.toSimpleString() }}) -> ${this.returnType}"
                 is Function.Parameter -> "${this.name}: ${this.type}${if (this.defaultValue != null) " = ${this.defaultValue.toSimpleString()}" else ""}"
                 is Expression.FunctionCall.Argument -> "${if (this.name != null) "${this.name} = " else ""}${this.value.toSimpleString()}"
+
+                else -> {}
```
ignacy123, line 32: Are we sure this is desired? Because of this, adding a new type without addressing it in this `when` will no longer fail to compile.

app/src/main/kotlin/compiler/intermediate/ControlFlowGraphBuilder.kt
```diff
@@ -1,41 +1,42 @@
...
 
 class ControlFlowGraphBuilder(@JvmField var entryTreeRoot: IFTNode? = null) {
-    private var unconditionalLinks = referenceHashMapOf<IFTNode, IFTNode>()
-    private var conditionalTrueLinks = referenceHashMapOf<IFTNode, IFTNode>()
-    private var conditionalFalseLinks = referenceHashMapOf<IFTNode, IFTNode>()
-    private var treeRoots = referenceHashSetOf<IFTNode>()
+    private var unconditionalLinks = mutableRefMapOf<IFTNode, IFTNode>()
+    private var conditionalTrueLinks = mutableRefMapOf<IFTNode, IFTNode>()
+    private var conditionalFalseLinks = mutableRefMapOf<IFTNode, IFTNode>()
+    private var treeRoots = mutableListOf<Ref<IFTNode>>()
```
ignacy123, line 13: Any particular reasons for switching from set to list? We do perform lookups on this, so I think that the usage of the set was justified.

app/src/main/kotlin/compiler/lowlevel/dataflow/Liveness.kt
```diff
@@ -20,52 +16,52 @@ object Liveness {
...
-        val allRegisters = referenceHashSetOf<Register>()
+        val allRegisters = mutableSetOf<Register>()
         instructionList.forEach {
             allRegisters.addAll(it.regsDefined)
             allRegisters.addAll(it.regsUsed)
         }
 
-        val interferenceGraph = referenceHashMapOf(allRegisters.map { it to referenceHashSetOf<Register>() })
-        val copyGraph = referenceHashMapOf(allRegisters.map { it to referenceHashSetOf<Register>() })
+        val interferenceGraph = mutableMapOf(*allRegisters.map { it to mutableSetOf<Register>() }.toTypedArray())
```
ignacy123, line 47: Maybe we should have another map builder that will take a list explicitly? The conversion to an array followed by spreading is not readable.

app/src/test/kotlin/compiler/utils/TestUtils.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import kotlin.test.assertContains
+
+fun <E> assertContentEquals(expected: Set<E>, actual: Set<E>) {
+    for (element in expected)
+        assertContains(actual, element)
+    for (element in actual)
+        assertContains(expected, element)
+}
+
```
ignacy123, line 11: According to the [docs](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-content-equals.html), normal `assertEquals` should do exactly this. Or am I missing something? :thinking:

app/src/main/kotlin/compiler/lowlevel/linearization/IFTPattern.kt
```diff
@@ -51,8 +51,8 @@ sealed class IFTPattern {
 
     // Matches every intermediate form tree node
     // Doesn't consume the nodes, returns it as a subtree
-    class AnyNode : IFTPattern() {
-        override fun match(node: IFTNode): MatchResult? {
+    data class AnyNode(val dummy: Unit = Unit) : IFTPattern() {
```
ignacy123, line 54: Why the `dummy` arg?

ignacy123 (commented):
For such large refactors, I suggest splitting related changes into smaller commits as it is much easier to browse like that.
Good work on the actual changes, the codebase definitely needs this cleanup :+1:

app/src/main/kotlin/compiler/intermediate/ControlFlowGraphBuilder.kt
tomekb234: Isomorphism checking would be quite difficult without this.

app/src/main/kotlin/compiler/lowlevel/dataflow/Liveness.kt
tomekb234: This is only because Kotlin has no `toMutableMap` for some reason.

app/src/main/kotlin/compiler/utils/Ref.kt
tomekb234: Good idea, actually. Removed.

app/src/main/kotlin/compiler/ast/AstNode.kt
tomekb234: So, I did two things: I made `Program` an `AstNode` and replaced `Any`s with `AstNode`s in some places. Since there was case checking on `Any` (!) and it became case checking on `AstNode`, Kotlin *then* started complaining that the checks are not exhaustive, so I added these `else -> {}`, but I got carried away and also added them in `AstNode`.

tomekb234: Fixed.

app/src/main/kotlin/compiler/intermediate/ControlFlowGraphBuilder.kt
tomekb234: Basically, the output order of nodes in `ControlFlowGraphBuilder` must be deterministic. I added an optimization to use both a list (for storing order) and a set (for lookups).

app/src/main/kotlin/compiler/lowlevel/linearization/IFTPattern.kt
tomekb234: Because we do not want bare `class`es when other cases in a `sealed class` are `data class`es, and a `data class` needs at least one field. This trick satisfies this requirement.

app/src/test/kotlin/compiler/utils/TestUtils.kt
tomekb234: That is right. I though that `assertEquals` also checks if the types are the same. We didn't actually need Kotlin's `assertContentEquals` at all, so I replaced all its uses with `assertEquals`.

app/src/main/kotlin/compiler/lowlevel/linearization/IFTPattern.kt
tomekb234: Also, replacing such cases with `object`s is generally not the best idea, because sometimes we want to distinguish by identity. In case of `IFTPattern` this shouldn't be a problem (I only changed this for consistency), because we do not use it with `Ref`, but for example `IFTNode.Dummy` definitely should not be an `object`.

ignacy123 (approved):

PR #139, opened 2023-01-03
Generator functions template

ggawryal:


app/src/main/kotlin/compiler/ast/Function.kt
```diff
@@ -2,13 +2,7 @@ package compiler.ast
...
 import compiler.input.LocationRange
 
-data class Function(
-    override val name: String,
-    val parameters: List<Parameter>,
-    val returnType: Type,
-    val implementation: Implementation,
-    override val location: LocationRange? = null,
-) : NamedNode, AstNode, VariableOwner {
+sealed class MetaFunction : NamedNode, AstNode, VariableOwner {
```
tomekb234, line 5: Wouldn't it be easier to just add `isGenerator: Boolean` to `Function`? The field `returnType` could then mean either "return type" or "yield type" (which can be thought of as "generalized return"). I think functions and generator functions will actually have quite a lot in common, and this separation may lead to large amount of code duplication.

tomekb234: Okay, maybe there won't be too much code duplication if we have this `MetaFunction` class, but still, since `Function` and `Generator` are basically the same thing, the difference being only the type and the interpretation of `returnType`, instead of writing `f is Generator` we may as well write `f.isGenerator`. This would be more consistent with the fact that foreign and local functions have the same type, despite being somewhat different things.

tomekb234:
We also need the three methods `genInit`, `genContinue` and `genBreak`, either in separate `GeneratorDetailsGenerator` (or something with less confusing name), or in `FunctionDetailsGenerator` , depending on whether we choose to consider generator functions as something completely different than functions, or just as special cases of functions which are called differently and whose results are retrieved differently.

I would favor the second option (generator functions still being `Function`s), because I suspect that even if `FunctionDetailsGenerator`s were only dedicated for regular functions, they would still need to be aware of the existence of generator functions, because they should probably behave differently when generating intermediate code for regular function calls inside the bodies of generator functions, and so we may as well integrate the code for both regular functions and generator functions inside `FunctionDetailsGenerator`.

app/src/main/kotlin/compiler/ast/Function.kt
ggawryal: Thought that using `MetaFunction` would be more convenient, because it could help find `when` clauses where we'll be missing generator handling, but I agree that it would have a lot in common with normal `Function`s and we should be consistent with `Foreign` functions. Refactored as you suggested.

app/src/main/kotlin/compiler/ast/Statement.kt
```diff
@@ -52,6 +52,13 @@ sealed class Statement : AstNode {
         override val location: LocationRange? = null,
     ) : Statement()
 
+    data class ForeachLoop(
+        val receivingVariable: Variable,
```
tomekb234, line 56: This should be `String`, but I guess in this way the implementation will be somewhat easier.

app/src/main/kotlin/compiler/intermediate/generators/FunctionDetailsGenerator.kt
```diff
@@ -10,12 +10,18 @@ interface FunctionDetailsGenerator : VariableAccessGenerator {
         val result: IFTNode?
```
tomekb234, line 10: The calls which resume a generator function will almost certainly return two values (the state and the passed value), so we may as well already add `val secondResult: IFTNode? = null` here.

tomekb234:
An AST node type the yield statement is missing. `FunctionDetailsGenerator` should also have `fun genYield(value: IFTNode): ControlFlowGraph`.

app/src/main/kotlin/compiler/intermediate/generators/FunctionDetailsGenerator.kt
```diff
@@ -10,12 +10,18 @@ interface FunctionDetailsGenerator : VariableAccessGenerator {
...
     fun genCall(args: List<IFTNode>): FunctionCallIntermediateForm
 
+    // normal functions, callee-side
     fun genPrologue(): ControlFlowGraph
-
     fun genEpilogue(): ControlFlowGraph
 
+    // generators, callee-side
+    fun genInit(): ControlFlowGraph
+    fun genResume(): ControlFlowGraph
```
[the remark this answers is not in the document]
tomekb234: And just to clarify: The prologue and epilogue will be used to build the resume function, which will contain the generator's actual code. So, basically, what I mean is that we should split `genResume` into `genResumePrologue` and `genResumeEpilogue`, but since these two will be very similar to `genPrologue` and `genEpilogue` of normal functions, we may as well reuse the latter and only add checks for `isGenerator` in them.

app/src/main/kotlin/compiler/intermediate/generators/FunctionDetailsGenerator.kt
ggawryal: True, but I'm not sure if `IFTNode` would be sufficient or should we use a `ControlFlowGraph` for storing the state - it should restore potentially large memory state on heap and set `RIP` to appropriate address - even if we add some `memcpy` as `IFTNode`, these are at least two separate operations

app/src/main/kotlin/compiler/intermediate/generators/FunctionDetailsGenerator.kt
```diff
@@ -7,20 +7,20 @@ import compiler.intermediate.IFTNode
 interface FunctionDetailsGenerator : VariableAccessGenerator {
     data class FunctionCallIntermediateForm(
         val callGraph: ControlFlowGraph,
-        val result: IFTNode?
+        val result: IFTNode?,
+        val stateForResumingGenerator: IFTNode?
```
tomekb234, line 11: This can actually be general-purpose (SysV calling convention allows using both RAX and RDX for return values), that's why I proposed to simply name it `secondResult`. Currently we are only planning to use it for generator functions, so I guess it's not an issue for now.

tomekb234 (approved):

app/src/main/kotlin/compiler/intermediate/generators/FunctionDetailsGenerator.kt
tomekb234: Copying won't be necessary. I posted the details on our group chat.

tomekb234 (approved):

PR #140, opened 2023-01-03
Check whether all accessed variables are initialized

ignacy123:
The check is run for all top-level functions, even the unused ones (assuming that `główna` is the only entry point). Some tests rely on this assumption, so this ensures compatibility.

### DIFF
--- a/app/src/main/kotlin/compiler/analysis/InitializationVerifier.kt
+++ b/app/src/main/kotlin/compiler/analysis/InitializationVerifier.kt
@@ -70,21 +70,15 @@ object InitializationVerifier {
                 is Statement.Evaluation -> verifyInitialization(node.expression, initializedVariables, traversedFunctions)
                 is Statement.VariableDefinition -> verifyInitialization(node.variable, initializedVariables, traversedFunctions)
                 is Global.VariableDefinition -> verifyInitialization(node.variable, initializedVariables, traversedFunctions)
-                is Statement.FunctionDefinition -> {
-                    verifyInitialization(node.function, initializedVariables, traversedFunctions)
-                }
+                is Statement.FunctionDefinition -> verifyInitialization(node.function, initializedVariables, traversedFunctions)
                 is Global.FunctionDefinition -> verifyInitialization(node.function, initializedVariables, traversedFunctions)
                 is Statement.Assignment -> {
                     verifyInitialization(node.value, initializedVariables, traversedFunctions)
                     // the variable will be initialized after assignment, not before
                     initializedVariables.add(nameResolution[Ref(node)]!!)
                 }
-                is Statement.Block -> {
-                    node.block.forEach { verifyInitialization(it, initializedVariables, traversedFunctions) }
-                }
-                is Statement.Conditional -> {
-                    handleConditional(node.condition, node.actionWhenTrue, node.actionWhenFalse)
-                }
+                is Statement.Block -> node.block.forEach { verifyInitialization(it, initializedVariables, traversedFunctions) }
+                is Statement.Conditional -> handleConditional(node.condition, node.actionWhenTrue, node.actionWhenFalse)
                 is Statement.Loop -> {
                     verifyInitialization(node.condition, initializedVariables, traversedFunctions)
                     handleIsolatedStatementBlock(node.action)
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/compiler/analysis/InitializationVerifier.kt
```diff
@@ -0,0 +1,155 @@
...
+                    node.actionWhenFalse?.forEach {
+                        verifyInitialization(
+                            it, initializedVariablesWhenFalse,
+                            traversedFunctionsWhenFalse
+                        )
+                    }
+
+                    val initializedVariablesWhenBoth = initializedVariablesWhenTrue intersect initializedVariablesWhenFalse
+                    val traversedFunctionsWhenBoth = traversedFunctionsWhenTrue intersect traversedFunctionsWhenFalse
+                    initializedVariables.clear()
```
ggawryal, line 66: If I'm not mistaken, these `clear`s are not needed - if so, why don't we remove them?

ggawryal (changes requested):
There is a problem with `oraz` and `lub` expressions and their short circuit semantics. Except of this code looks good.

ignacy123:
I added proper handling of returns, as I forgot about it before :disappointed:

app/src/main/kotlin/compiler/analysis/InitializationVerifier.kt
```diff
@@ -0,0 +1,200 @@
...
+        ast: Program,
+        nameResolution: Map<Ref<AstNode>, Ref<NamedNode>>,
+        defaultParameterMapping: Map<Ref<Function.Parameter>, Variable>,
+        diagnostics: Diagnostics
+    ) {
+        fun verifyInitialization(
+            node: AstNode,
+            state: VerificationState
+        ): VerificationState? {
+            // Consider a function f, which performs g(). g might contain conditional returns.
```
ggawryal, line 50: Great catch with these conditional returns. I think we should also treat `yield` as a possible conditional return, since `foreach` loop can be exited using `break` statement.

app/src/test/kotlin/compiler/analysis/InitializationVerifierTest.kt
```diff
@@ -0,0 +1,934 @@
...
+        val program = Program(listOf(FunctionDefinition(function)))
+        val nameResolution: Map<Ref<AstNode>, Ref<NamedNode>> = refMapOf(
+            assignmentToX to x,
+            readFromX to x
+        )
+
+        checkDiagnostics(program, nameResolution, listOf(ReferenceToUninitializedVariable(x)))
+    }
+
+// czynność zewnętrzna() {
```
ggawryal, line 171:
```suggestion
    // czynność zewnętrzna() {
```

app/src/test/kotlin/compiler/analysis/InitializationVerifierTest.kt
```diff
@@ -0,0 +1,934 @@
...
+        val nameResolution: Map<Ref<AstNode>, Ref<NamedNode>> = refMapOf(
+            readFromX to x
+        )
+
+        checkDiagnostics(program, nameResolution, listOf(ReferenceToUninitializedVariable(x)))
+    }
+
+    // czynność zewnętrzna() {
+    //     zm x: Liczba
+    //     jeżeli (fałsz) {
```
ggawryal, line 204:
```suggestion
    //     jeśli (fałsz) {
```

app/src/test/kotlin/compiler/analysis/InitializationVerifierTest.kt
```diff
@@ -0,0 +1,934 @@
...
+        checkDiagnostics(program, nameResolution, listOf())
+    }
+
+    // czynność zewnętrzna() {
+    //     zm x: Liczba
+    //     czynność wewnętrzna() -> Czy {
+    //          x = 123
+    //          zwróć fałsz
+    //     }
+    //     jeżeli (wewnętrzna()) {
```
ggawryal, line 246:
```suggestion
    //     jeśli (wewnętrzna()) {
```

ggawryal (changes requested):

ggawryal (approved):
LGTM

PR #141, opened 2023-01-06
Yield syntax

JHodorHodor:


app/src/test/kotlin/compiler/e2e/CorrectProgramsE2eTest.kt
```diff
@@ -215,4 +215,37 @@ class CorrectProgramsE2eTest {
...
+            czynność główna() {
+                przekaźnik mój_generator(a: Liczba = 17) -> Liczba {
+                    zm x: Liczba = 1
+                    dopóki (x <= a) {
+                        przekaż x
+                
+                        x = x + 1
+                
+                        if (x == 42)
+                            zakończ // możliwa tylko forma bez wartości
```
ggawryal, line 233: Since it's e2e test, how about adding another `przekaż` to that function, for example there? It would be useful later for testing if the function returns control back to the correct yield statement (out of 2)?

app/src/test/kotlin/compiler/syntax/AstFactoryTest.kt
```diff
@@ -228,6 +281,28 @@ class AstFactoryTest {
                 makeTNode(TokenType.IDENTIFIER, "poboczna3")
             ),
             makeTNode(TokenType.NEWLINE, "\n"),
+            makeNTNode(
+                NonTerminalType.FOREIGN_DECL, Productions.foreignDecl,
+                makeTNode(TokenType.FOREIGN, "zewnętrzna"),
```
ggawryal, line 286: "Zewnętrzna przekaźnik" sounds odd. Guess we wouldn't have many foreign generators, so it's not a big issue, but I think it is a place for improvements, possibly in next sprints.

app/src/main/kotlin/compiler/syntax/AstFactory.kt
```diff
@@ -360,6 +360,22 @@ object AstFactory {
...
+                    Variable.Kind.VALUE,
+                    (children[1] as ParseTree.Leaf).content,
+                    processType(children[3]),
+                    null,
+                    combineLocations(children.subList(1, 3)),
+                )
+                val generatorCall = Expression.FunctionCall(
+                    (children[5] as ParseTree.Leaf).content,
+                    processFunctionCallArguments(children[7], diagnostics),
+                    combineLocations(children.subList(5, 7)),
```
ggawryal, line 374: Could you double check if locations are computed correctly? In Kotlin, `toIndex` in `List.subList` is exclusive and we also need to account closing brace, so imho it should be `(5,9)`.

ggawryal (approved):
LGTM, left only a few not very important comments.

bebidek:
I would suggest also updating VIM syntax files to include new keywords.

app/src/test/kotlin/compiler/syntax/AstFactoryTest.kt
JHodorHodor: What do you think about changing to `zew`?

app/src/main/kotlin/compiler/syntax/AstFactory.kt
JHodorHodor: You're right

app/src/test/kotlin/compiler/syntax/AstFactoryTest.kt
JHodorHodor: For now, I changed to `zew`, let me know what do you think.

tomekb234: We can just add another keyword `zewnętrzny`.

ggawryal: For me we `zew` is good enough, or we can add `zewnętrzny` with grammar rules so that it matches only `przekaźnik` (and `zewnętrzna` only `czynność`).

JHodorHodor: If I was a user of this language then I would really prefer not to be able to make such “hard to detect” errors. So I would go with “zew” unless someone strongly opposes.

tomekb234: I would just add `zewnętrzny` as a synonym for `zewnętrzna` (both usable before `czynność` and `przekaźnik`) so that the user could write it nicely, but also would not have to bother with those "hard to detect" errors. If you really want a single keyword for it, then I would suggest to change it to more suggestive `zewn`.

JHodorHodor: This sounds like a nice solution, I will add two options.

tomekb234: (Jakże ważna dyskusja się tu zrobiła xd)

PR #142, opened 2023-01-07
Add CFG generation for foreach and yield

kamgal00:


app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
```diff
@@ -65,7 +81,7 @@ object ControlFlow {
                 functionDetailsGenerators[function]!!.genPrologue(),
                 functionDetailsGenerators[function]!!.genEpilogue()
             )
-        }
+        } // TODO: add initialization and finalization of generators
```
ggawryal, line 84: Am I right, that this comment is obsolete, since generators are initialized and finalized in foreach loops?

app/src/test/kotlin/compiler/intermediate/FunctionControlFlowTest.kt
```diff
@@ -49,7 +49,9 @@ class FunctionControlFlowTest {
         object : Diagnostics {
             override fun report(diagnostic: Diagnostic) { diagnostics.add(diagnostic) }
             override fun hasAnyError(): Boolean { throw RuntimeException("This method shouldn't be called") }
-        }
+        },
+        { _, _, _ -> throw NotImplementedError() }, // TODO: write some tests for yield and foreach
```
ggawryal, line 53: Those test would be quite useful indeed.

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
```diff
@@ -466,13 +526,77 @@ object ControlFlow {
...
+                            returning.addAll(last)
                             last = emptyList()
                         }
 
-                        is Statement.ForeachLoop -> TODO()
-                        is Statement.GeneratorYield -> TODO()
+                        is Statement.ForeachLoop -> {
+                            val gdg = getGeneratorDetailsGenerator(nameResolution[Ref(statement.generatorCall)]!!.value as Function)
+                            val frameMemoryAddress = if (function.isGenerator)
+                                getGeneratorDetailsGenerator(function).getNestedForeachFramePointerAddress(statement)
```
ggawryal, line 536: If i understand correctly, we cannot use `getNestedForeachFramePointerAddress(statement)` there, because foreign generators don't have it implemented. I think we should check if the implementation of that generator is local, and if it's not, then `frameMemoryAddress=null` should be imo ok.

ggawryal (changes requested):
If that's not a problem, we can add some more code to make foreign function details generator fully working.

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
kamgal00: We still need to add somewhere the code responsible for initialization, resume, and finalization of generators, I mean the code returned by  genInit, genResume and genFinalize in GeneratorDetailsGenerator

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
kamgal00: Actually, this line is OK. `function` is given as a parameter in `processFunction` and thus have to be local.

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
ggawryal: Got it, so we should probably create separate maps for `initialization` and `finalization` and pass them to linearization phase. We can left the integration for another task if you feel this is out of scope for this task.

kamgal00: Yeah, I think we can do this in the next week

app/src/main/kotlin/compiler/intermediate/FunctionDependenciesAnalyzer.kt
```diff
@@ -87,39 +89,51 @@ object FunctionDependenciesAnalyzer {
         variableProperties: Map<Ref<AstNode>, VariablePropertiesAnalyzer.VariableProperties>,
         functionReturnedValueVariables: Map<Ref<Function>, Variable>,
         allowInconsistentNamingErrors: Boolean = false
-    ): Map<Ref<Function>, FunctionDetailsGenerator> {
-        val result = mutableKeyRefMapOf<Function, FunctionDetailsGenerator>()
+    ): Pair<Map<Ref<Function>, FunctionDetailsGenerator>, Map<Ref<Function>, GeneratorDetailsGenerator>> {
```
ggawryal, line 92: Maybe rename this method to `createCallablesDetailsGenerator`

app/src/test/kotlin/compiler/intermediate/ControlFlowComparisonUtils.kt
```diff
@@ -0,0 +1,144 @@
...
+    val nodeMap = mutableKeyRefMapOf<IFTNode, IFTNode>()
+
+    fun <T> MutableMap<Ref<T>, T>.ensurePairSymmetrical(a: T, b: T) {
+        if (!this.containsKey(Ref(a))) {
+            this[Ref(a)] = b
+        }
+        assertSame(this[Ref(a)]!!, b)
+    }
+
+    infix fun IFTNode.hasSameStructureAs(iftNode: IFTNode) {
```
ggawryal, line 23: Since this method now does assertions instead of returning bool, maybe rename it to `assertHasSameStructureAs` or `hasSameStructureAssertedAs`

app/src/test/kotlin/compiler/intermediate/FunctionControlFlowTest.kt
```diff
@@ -601,4 +674,254 @@ class FunctionControlFlowTest {
 
         assertEquals(listOf<Diagnostic>(ControlFlowDiagnostic.Warnings.UnreachableStatement(evaluation)), diagnostics)
     }
+
+    @Test
```
ggawryal, line 678: For readability and to keep consistency with other test, we can add in comments how does the code look like, eg.
```
przekaźnik generator() {}
f() {
    otrzymując x: Liczba od generator() {123}
}
```

ggawryal (approved):
Loos good, left just a few minor comments

app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -56,16 +56,20 @@ class Compiler(val diagnostics: Diagnostics) {
 
             val programProperties = ProgramAnalyzer.analyzeProgram(astWithBuiltinFunctions, diagnostics)
 
-            val functionDetailsGenerators = FunctionDependenciesAnalyzer.createFunctionDetailsGenerators(
+            val (functionDetailsGenerators, generatorDetailsGenerators) = FunctionDependenciesAnalyzer.createFunctionDetailsGenerators(
```
ggawryal, line 59: We can also add foreign generators `init`s, `resume`s and `finalize`s to foreign identifiers in line 106, similarly to functionDetailsGenerators, and remove that todo.

ggawryal (approved):
LGTM

PR #143, opened 2023-01-07
Name res generators

woocash2:


app/src/test/kotlin/compiler/analysis/NameResolverTest.kt
```diff
@@ -104,6 +105,124 @@ internal class NameResolverTest {
         assertEquals(1, result.programStaticDepth)
     }
 
+    @Test fun `test name resolution with generator`() {
+        /*
+        Create AST for program:
+        ---------------------------------
+        zew czynność f(a: Liczba) -> Liczba {
```
ggawryal, line 112:
```suggestion
        przekaźnik f(a: Liczba) -> Liczba {
```

app/src/test/kotlin/compiler/analysis/NameResolverTest.kt
```diff
@@ -86,6 +86,7 @@ internal class NameResolverTest {
         val nameDefinitions = result.nameDefinitions
 
         // Check the result mapping
+        assertEquals(5, nameDefinitions.size)
```
ggawryal, line 89: We should remove a duplicated assert in line `97`

app/src/main/kotlin/compiler/analysis/NameResolver.kt
```diff
@@ -69,7 +73,7 @@ object NameResolver {
                     diagnostics.report(Diagnostic.ResolutionDiagnostic.NameResolutionError.UndefinedVariable(astNode))
                 if (astNode is Statement.Assignment)
                     diagnostics.report(Diagnostic.ResolutionDiagnostic.NameResolutionError.AssignmentToUndefinedVariable(astNode))
-            } else if (visibleNames[name]!!.onlyHasFunctions()) {
+            } else if (visibleNames[name]!!.onlyHasCallables()) {
                 if (astNode is Expression.Variable)
                     diagnostics.report(Diagnostic.ResolutionDiagnostic.NameResolutionError.FunctionIsNotVariable(visibleNames[name]!!.topFunction(), astNode))
```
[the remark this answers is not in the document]
ggawryal: Also we could possibly add an unit test for that to make sure it fails with appropriate diagnostic

ggawryal (changes requested):
The idea to separate `generators` stack from other is correct, but I think there could be some bugs. Please read comments and double check that.

app/src/main/kotlin/compiler/analysis/NameResolver.kt
```diff
@@ -196,7 +204,9 @@ object NameResolver {
 
                 is Expression.FunctionCall -> {
                     if (!checkFunctionUsage(node)) {
-                        nameDefinitions[Ref(node)] = Ref(visibleNames[node.name]!!.topFunction())
+                        // if this is a generator call then analysis of Statement.ForEachLoop has already resolved this name
```
ggawryal, line 207: Also please watch out there for incorrectly calling generator as it was a normal function, since previous phases don't and cannot detect it.

app/src/main/kotlin/compiler/analysis/NameResolver.kt
```diff
@@ -270,6 +280,28 @@ object NameResolver {
...
                 }
+
+                is Statement.ForeachLoop -> {
+                    val newScope = makeScope() // foreach body introduces a new scope
+
+                    // receiving variable and generator call are treated as they belonged to the new scope
+                    analyzeNode(node.receivingVariable, newScope)
+
+                    // resolve name for generator call. The subsequent analyzeNode won't update the name resolution for this call
+                    if (!checkFunctionUsage(node.generatorCall)) {
```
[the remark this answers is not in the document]
ggawryal: In such case we can return `FunctionUsedAsAGeneratorError` or sth similar

app/src/main/kotlin/compiler/analysis/NameResolver.kt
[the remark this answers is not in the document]
woocash2: True

app/src/main/kotlin/compiler/analysis/NameResolver.kt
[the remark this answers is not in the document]
woocash2: Good point

app/src/main/kotlin/compiler/analysis/NameResolver.kt
woocash2: Now I changed it so this case will happen only if this is a regular function call.
Generator calls can only occur in `ForEachLoop` which will perform name resolution and **won't** recursively analyse the `generatorCall`, but directly analyse its arguments.

app/src/main/kotlin/compiler/analysis/NameResolver.kt
```diff
@@ -69,11 +75,29 @@ object NameResolver {
...
-                if (astNode is Statement.Assignment)
-                    diagnostics.report(Diagnostic.ResolutionDiagnostic.NameResolutionError.AssignmentToFunction(visibleNames[name]!!.topFunction(), astNode))
+            } else if (!visibleNames[name]!!.hasVariable()) { // then it must have at least one function or generator
+                if (astNode is Expression.Variable) {
+                    diagnostics.report(
+                        Diagnostic.ResolutionDiagnostic.NameResolutionError.CallableIsNotVariable(
+                            if (visibleNames[name]!!.hasFunction())
+                                visibleNames[name]!!.topFunction()
+                            else
+                                visibleNames[name]!!.topGenerator(),
```
ggawryal, line 85: We can extract this common expression

ggawryal (approved):
LGTM 👍

PR #144, opened 2023-01-07
Add simple generators to stdlib

tomekb234:


app/src/main/kotlin/compiler/analysis/BuiltinFunctions.kt
```diff
@@ -13,12 +13,29 @@ object BuiltinFunctions {
...
+
         "wczytaj" to Function(
             "wczytaj",
             emptyList(),
             Type.Number,
             Function.Implementation.Foreign("read_int64"),
             false
+        ),
+
+        "przedziału" to Function(
```
ggawryal, line 25: What do you think of adding some documentation there? It's not obvious without reading C code or unit tests whether `max` is inclusive or not and what is the starting value.

ggawryal (approved):
Looks good!

app/src/main/kotlin/compiler/analysis/BuiltinFunctions.kt
tomekb234: Added comments to `BuiltinFunctions`. The documentation for this language (including the standard functions) should be in some separate document, anyway.

PR #145, opened 2023-01-07
Add type checking for generators

rkilar:


app/src/main/kotlin/compiler/analysis/TypeChecker.kt
```diff
@@ -137,7 +148,7 @@ class TypeChecker(private val nameResolution: Map<Ref<AstNode>, Ref<NamedNode>>,
 
         checkBlock(function.body)
 
-        if (function.implementation is Function.Implementation.Local && function.returnType != Type.Unit)
+        if (function.implementation is Function.Implementation.Local && function.returnType != Type.Unit && !function.isGenerator)
```
ggawryal, line 151: Similar extra condition (`&& !function.isGenerator`) will be needed in `compiler.analysis.ReturnValueVariableCreator:17`

ggawryal (changes requested):
Generally looks good, but there is one issue which we should take care of either there of in parser + 2 comments, one minor and one more important.

ggawryal (approved):
LGTM 👍

PR #146, opened 2023-01-07
Generator details generator

bebidek:


app/src/main/kotlin/compiler/intermediate/generators/DefaultGeneratorDetailsGenerator.kt
```diff
@@ -0,0 +1,250 @@
...
+        getNestedForeachFramePointerAddressWithCustomBase(foreachLoop, Register.RBP)
+
+    override fun genFinalize(): ControlFlowGraph {
+        val cfgBuilder = ControlFlowGraphBuilder()
+
+        // generate the prologue
+        cfgBuilder.mergeUnconditionally(finalizeFDG.genPrologue())
+
+        // prepare custom base
+        val customBaseRegister = Register.R15 // R15 is callee-saved FIXME: consider using a virtual register here
```
ggawryal, line 177: To improve this, we'd need to modify coloring to allow it picking arbitrary `calee-saved` register there, right? We can create an issue for that after merging.

app/src/main/kotlin/compiler/intermediate/generators/DefaultGeneratorDetailsGenerator.kt
```diff
@@ -0,0 +1,250 @@
...
+class DefaultGeneratorDetailsGenerator(
+    private val parameters: List<NamedNode>,
+    initLabel: IFTNode.MemoryLabel,
+    private val resumeLabel: IFTNode.MemoryLabel,
+    finalizeLabel: IFTNode.MemoryLabel,
+    depth: ULong,
+    variablesLocationTypes: Map<Ref<NamedNode>, VariableLocationType>,
+    displayAddress: IFTNode,
+    private val nestedForeachLoops: List<Ref<Statement.ForeachLoop>>,
+    private val nameResolution: Map<Ref<AstNode>, Ref<NamedNode>>, // constructor doesn't use this argument, thus incomplete map can be passed
```
ggawryal, line 26: These two maps are only used in one place. Maybe it would be easier to pass instead a function that for a given foreach loop in `nestedForeachLoops` returns `GeneratorDetailsGenerator`

app/src/main/kotlin/compiler/intermediate/generators/GeneratorDetailsGenerator.kt
```diff
@@ -0,0 +1,29 @@
+package compiler.intermediate.generators
+
+import compiler.ast.Statement
+import compiler.intermediate.ControlFlowGraph
+import compiler.intermediate.IFTNode
+
+interface GeneratorDetailsGenerator : VariableAccessGenerator {
```
ggawryal, line 7: I'm not a huge fan of this name, but we can leave it as some kind of a joke, or change it to `GeneratorDetailsCreator`.

ggawryal (approved):
Great contribution, thanks. I think we were supposed to do only `ForeignGeneratorDetailsGenerator`, but the implementation of `DefaultGeneratorDetailsGenerator` visually and conceptually also looks correct.

app/src/main/kotlin/compiler/intermediate/generators/GeneratorDetailsGenerator.kt
bebidek: Actually, I really like that name (also, it is consistent with `FunctionDetailsGenerator`), so I'll keep it.

app/src/main/kotlin/compiler/intermediate/generators/ForeignGeneratorDetailsGenerator.kt
```diff
@@ -0,0 +1,42 @@
...
+        return finalizeFDG.genCall(listOf(framePointer))
+    }
+
+    override fun genInit() = throw NotImplementedError()
+
+    override fun genResume(mainBody: ControlFlowGraph) = throw NotImplementedError()
+
+    override fun genYield(value: IFTNode) = throw NotImplementedError()
+
+    override fun getNestedForeachFramePointerAddress(foreachLoop: Statement.ForeachLoop) = throw NotImplementedError()
```
ggawryal, line 35: This method will be used by AST - > CFG, so maybe return `null` instead throwing an `NotImplementedError`, or some more specific expection?

stdlib/stdlib.c
```diff
@@ -10,3 +15,13 @@ int64_t read_int64() {
     scanf("%" SCNd64, &value);
     return value;
 }
+
+void* _$checked_malloc(size_t size) {
+    void* address = malloc(size);
+    if (size > 0 && address == NULL) {
+        fprintf(stderr, "%s\n", strerror(ENOMEM));
+        exit(1);
```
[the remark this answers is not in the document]
bebidek: Process exit codes and errno codes are two different namespaces and returning errno code is not an usual practice. For example, `cat` command on my Linux machine return exit code 1 for both "No such file or directory" and "Permission denied" error, even though `ENOENT` and `EPERM` have different values.

app/src/main/kotlin/compiler/intermediate/generators/DefaultGeneratorDetailsGenerator.kt
bebidek: I think it could work without modification in coloring. This register will be in conflict with all caller-saved ones (because of CALL's), so the allocator will either give it callee-saved register (which is fine) or spill it (which should also be fine I guess). But I'm not sure about that and I can't test this class at the moment, so I did it a safer way.

app/src/main/kotlin/compiler/intermediate/generators/ForeignGeneratorDetailsGenerator.kt
bebidek: No, it won't be used there. Both `genYield` and `getNestedForeachFramePointerAddress` will be called while processing generator's body, but foreign generators don't have a body (I mean - in the code we compile).

app/src/main/kotlin/compiler/intermediate/IFTNode.kt
```diff
@@ -51,6 +54,7 @@ sealed class IFTNode {
     data class StackPush(val node: IFTNode) : IFTNode()
     data class StackPop(val dummy: Unit = Unit) : IFTNode()
 
+    data class JumpToRegister(val targetRegister: Register) : IFTNode()
```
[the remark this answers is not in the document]
bebidek: Then I should add it to `Return` as well.

PR #155, opened 2023-01-10
Add interfaces for arrays

ignacy123:


app/src/main/kotlin/compiler/ast/Expression.kt
```diff
@@ -23,6 +23,17 @@ sealed class Expression : AstNode {
         override val location: LocationRange? = null,
     ) : Expression()
 
+    data class ArrayElement(
+        val name: String,
```
bebidek, line 27: Shouldn't we have an `Expression` here (since we can return an array from a function or have array of an array).
(also, we have `Expression` in `ArrayLength`)

app/src/main/kotlin/compiler/analysis/TypeChecker.kt
```diff
@@ -173,6 +173,8 @@ class TypeChecker(private val nameResolution: Map<Ref<AstNode>, Ref<NamedNode>>,
                         is Function -> report(TypeCheckingError.FunctionAsValue(expression, node))
                     }
                 }
+                is Expression.ArrayElement -> TODO()
+                is Expression.ArrayLength -> TODO()
```
bebidek, line 177: If we add TODOs here, then we should also add them in other analysis classes (e.g. NameResolver, ArgumentResolver...).

app/src/main/kotlin/compiler/ast/AstNode.kt
```diff
@@ -58,7 +60,14 @@ sealed interface AstNode {
                 is Expression -> "expression << ${this.toSimpleString()} >>"
 
                 is Statement -> when (this) {
-                    is Statement.Assignment -> "assignment << ${this.variableName} = ${this.value.toSimpleString()} >>"
+                    is Statement.Assignment -> {
+                        lvalue.let {
+                            when (it) {
+                                is Statement.Assignment.LValue.Variable -> "assignment << ${it.name} = ${this.value.toSimpleString()} >>"
+                                is Statement.Assignment.LValue.ArrayElement -> "assignment << ${it.name}[${it.index}] = ${this.value.toSimpleString()} >>"
```
bebidek, line 67: It would be more consistent to print `it.index.toSimpleString()` here.

app/src/main/kotlin/compiler/ast/Expression.kt
```diff
@@ -23,6 +23,24 @@ sealed class Expression : AstNode {
...
+    ) : Expression()
+
+    data class ArrayLength(
+        val expression: Expression,
+        override val location: LocationRange? = null
+    ) : Expression()
+
+    data class ArrayAllocation(
+        val name: String,
+        val size: Int,
```
bebidek, line 39: Do I understand correctly that this node is responsible for both (`arr = {2, 4, 3}` and `arr = new int[10]`) array creation methods?

bebidek (approved):

app/src/main/kotlin/compiler/ast/Expression.kt
ignacy123: Wasn't the syntax supposed to be `arr = new int[3](1, 2, 3)` (for the first case)? I remember we are against allowing default values, but I could be wrong.

tomekb234: It was supposed to be `arr = new int[3](1)`. The value 1 is copied 3 times.

PR #156, opened 2023-01-13
Initalize arrays during allocation with one value

ignacy123:


rkilar:
Is this actually how we want to do this? This allocation can be part of a complex expression. Right now the syntactic sugar would have to be translated in AstFactory, which would have to create new statements to realize this loop. Shouldn't this rather be done in the AST -> CFG step. I think this ast node should be left as it (apart from changing name to type and maybe adding some info which initialization method we are using).

ignacy123:
@woocash2 @kamgal00 Your thoughts?

kamgal00:
> Is this actually how we want to do this? This allocation can be part of a complex expression. Right now the syntactic sugar would have to be translated in AstFactory, which would have to create new statements to realize this loop. Shouldn't this rather be done in the AST -> CFG step. I think this ast node should be left as it (apart from changing name to type, size to an Expression and maybe adding some info which initialization method we are using). Also how do we handle allocation of multidimensional arrays using default values, don't we need a null/empty array literal?

I agree. Handling such nodes in AST->CFG is not a big problem, while it gives greater expressive power. I also believe that making size an Expression is a good idea.

app/src/main/kotlin/compiler/ast/Expression.kt
```diff
@@ -36,10 +36,16 @@ sealed class Expression : AstNode {
 
     data class ArrayAllocation(
         val name: String,
```
rkilar, line 38: Shouldn't this be a type not a name?

rkilar:
Also I don't remember if in the ended we wanted to have uninitialized arrays. Specifically, when we want to create an n by n 2D array did we want to allow `x = new [int][n](null); x[i] = new int[n](0)` or is `x = new [int][n](new int[n](0))` the only way (then the default value would have to be evaluated individually for each index so the arrays are different but we probably don't want that)?

ignacy123:
We actually only want the second type of initialization, with the default value being evaluated only once. A new 2d array should be an array of references to one array (evaluated during initialization).

rkilar (approved):

PR #157, opened 2023-01-13
Generators e2e tests

ggawryal:


tests/generators/local.tests
```diff
@@ -0,0 +1,46 @@
+// INTEGRATION TESTS IGNORE
```
ignacy123, line 1: This should be at the top of the file with code, not this one

tests/generators/stacked_foreach.bzz
```diff
@@ -0,0 +1,27 @@
...
+            jeśli(y > m) przerwij
+            otrzymując z: Liczba od przedziału(m+1) {
+                otrzymując w: Liczba od liczby_naturalne() {
+                    jeśli(w > m) przerwij
+                    wynik = wynik + x*y*w*(z+1)
+                }
+            }
+        }
+    }
+    //(1+2+3+..+m)^4
```
ignacy123, line 25: I didn't check the formula behind this code, but I assume you got it right :smile:

ignacy123 (changes requested):
Let's merge this after the other parts, so that the tests can pass

tests/generators/local_simple.bzz
```diff
@@ -0,0 +1,11 @@
+// INTEGRATION TESTS IGNORE
+przekaźnik p()-> Liczba {
+    przekaż 1
+}
+
+czynność główna() {
+    otrzymując x: Liczba od p(n)
```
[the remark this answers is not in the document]
ggawryal: Yeah, it should be simply `p()`

tests/generators/stacked_foreach.bzz
ggawryal: Actually, it's `1*nested_loop_result + 2*nested_loop_result + ... = (1+2+..+m)*nested_loop_result` :smiley:

tests/generators/mutual_recursion.bzz
```diff
@@ -0,0 +1,31 @@
...
+            zakończ
+        }
+        otrzymując x: Liczba od fib_przekaźnik(a-1) {
+            przekaż x+fib_czynność(a-2)
+            przerwij
+        }
+    }
+
+    zm wynik: Liczba = 0
+    otrzymując x: Liczba od fib_przekaźnik(a-1) {
```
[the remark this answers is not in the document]
ggawryal: You're right, it has been written incorrectly

ggawryal:
I've run test on #165 and they are passing. However, I found an unrelated bug with binary negation inside a conditional statement - its intermediate form looks correct, but asm code is wrong - I've added ignored e2e test for that and it's currently failing.

ignacy123 (approved):

PR #158, opened 2023-01-13
Properly handling end of life for variables

bebidek:
Currently a draft because assignments are not correctly handled (waiting for type checker PR).

### DIFF
--- a/app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
+++ b/app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
@@ -58,9 +58,14 @@ object ControlFlow {
             )
         }
 
+        val variableAccessGenerators = createVariableAccessGenerators(functionDetailsGenerators, generatorDetailsGenerators, globalVariableAccessGenerator)
+
+        fun createReadFromVariable(variable: Variable, currentFunction: Function): IFTNode {
+            val owner = programProperties.variableProperties[Ref(variable)]!!.owner
+            return variableAccessGenerators[Ref(owner)]!!.genRead(variable, owner === currentFunction)
+        }
+
         fun createWriteToVariable(iftNode: IFTNode, variable: Variable, currentFunction: Function): IFTNode {
-            val variableAccessGenerators: Map<Ref<VariableOwner>, VariableAccessGenerator> =
-                createVariableAccessGenerators(functionDetailsGenerators, generatorDetailsGenerators, globalVariableAccessGenerator)
             val owner = programProperties.variableProperties[Ref(variable)]!!.owner
             return variableAccessGenerators[Ref(owner)]!!.genWrite(variable, iftNode, owner === currentFunction)
         }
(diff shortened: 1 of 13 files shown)


bebidek:
This PR has most of commits from #163, so it should be merged after that one.

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
```diff
@@ -502,44 +536,75 @@ object ControlFlow {
...
+                    value.finalTreeRoots.forEach { cfgBuilder.addLink(it, treeHead.value, false) }
+                    treeHead.value = value.entryTreeRoot
+                }
+
+                fun addToDestructionStructures(variable: Variable) {
+                    val genDestructor = { variable.generateDestructor() }
+                    destructorsStack.add(genDestructor())
+                    addToEscapeTree(genDestructor(), returnTreeHead)
+                    addToEscapeTree(genDestructor(), breakTreeHead)
+                    addToEscapeTree(genDestructor(), continueTreeHead)
```
ignacy123, line 607: Why are we adding four different constructors for one variable? Can one variable be destructed multiple times, or are we ensuring that only of these will get called?

ignacy123 (changes requested):
From my understanding this PR is fixing some mistakes from `arrays-ast-to-cfg`. So let's merge this one into that branch first, and then everything into the `arrays` branch.

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
bebidek: I'm pretty sure you meant 'destructors'. And your second guess was correct, only one of them will be called (at least before variable is initialized again):

`destructorsStack` represents variables being destroyed because block they are defined in ends (destructors are inserted at the end of proper block).

Escape trees represent variables being destroyed because of `return`, `break` or `continue`. There is one returnEscapeTree for entire function and one break-/continueEscapeTree for each loop.

These four can't be merged (or at least I don't know how to do it in reasonable way), because even if, for example, `continue` and `break` would destroy the same variables, we need to end up in different place in code after that.

app/src/test/kotlin/compiler/intermediate/FunctionControlFlowTest.kt
```diff
@@ -807,36 +897,58 @@ class FunctionControlFlowTest {
...
-    //        }
-    //    }
+    // przekaźnik generator() {}
+    // f() {
+    //     otrzymując x: Liczba od generator() {
+    //         jeśli (prawda) {
+    //             123
+    //             przerwij
+    //         }
+    //         wpw {
```
ignacy123, line 941: wpp?

ignacy123 (changes requested):
LGTM :+1: 
Will approve after the comments on the base branch are addressed (so that diffs don't get mixed up).

bebidek:
> LGTM +1 Will approve after the comments on the base branch are addressed (so that diffs don't get mixed up).

Actually, I contacted with @kamgal00 earlier, so he probably waits for this PR merge.

ignacy123 (approved):

PR #160, opened 2023-01-14
Add arrays syntax

rkilar:
I noticed that the `type` field in `ArrayAllocation` was of type `String` instead of `Type`, which I missed when reviewing the PR.


app/src/test/kotlin/compiler/e2e/CorrectProgramsE2eTest.kt
```diff
@@ -249,4 +250,31 @@ class CorrectProgramsE2eTest {
             compiler.diagnostics.Diagnostic.ParserError::class,
         )
     }
+
+    @Ignore
+    @Test
+    fun `test arrays`() {
```
ignacy123, line 256: Please add two more tests before this - one per allocation type. It's gonna make it more obvious which allocation type is not implemented correctly during debugging

app/src/test/kotlin/compiler/syntax/AstFactoryTest.kt
```diff
@@ -692,20 +692,116 @@ class AstFactoryTest {
...
+                    dummyLocationRange
+                )
+            )
+        )
+
+        val resultAst = AstFactory.createFromParseTree(parseTree, dummyDiagnostics)
+        assertEquals(expectedAst, resultAst)
+    }
+
+    @Test fun `test correctly translates array allocation`() {
```
ignacy123, line 728: Please add a test for the one value allocation type

ignacy123 (changes requested):
We'll merge this after the other parts, but everything is looking good for me :+1:

app/src/main/kotlin/compiler/syntax/LanguageTokens.kt
```diff
@@ -34,6 +36,10 @@ object LanguageTokens {
             Pair(TokenType.FOREIGN2, "zewnętrzny"),
             Pair(TokenType.AS, "jako"),
 
+            // Array related keywords
+            Pair(TokenType.LENGTH, "długość"),
+            Pair(TokenType.ALLOCATE, "alokacja"),
```
bebidek, line 41: Are you sure that "alokacja" is the best choice here? While it is a Polish word, it's a loan translation from English "allocation" while we tend to prefer words of more Polish origin (look for example at "czynność" instead of "funkcja").

rkilar: Yeah, I wasn't so sure about it myself. I used it to match the string representation in ASTNode. I think "przydział" is the best Polish alternative. We can also use word not related to "allocate" but maybe "new", "create", "array" or "table". I need some outside opinion on that.

ignacy123: Maybe just "tablica"?

bebidek: Or "ciąg" if we prefer unusual keywords.

rkilar: It could also be "rząd", "szyk" (this one technically has German etymology) or "szereg" (from Hungarian).

ignacy123 (approved):

rkilar:
I had some problems with missing cases in type checking. I think I fixed them, as the end to end tests now fail on unimplemented cases in InitializationVerifier, but it should probably be checked.

ignacy123 (changes requested):
I also came across these problems, and already prepared a [new branch](https://github.com/TCS-Compilers/compiler/tree/fix-type-checking) with fixes.
Since you already started applying the fixes here, can you unify my work and yours before merging? I found one more issue in name resolution and added a new test to type checking (though you will have to adjust the final assert, as you are raising a different exception).

ignacy123 (approved):

PR #162, opened 2023-01-14
For each analyzer

JHodorHodor:


app/src/main/kotlin/compiler/analysis/GeneratorResolver.kt
```diff
@@ -0,0 +1,62 @@
...
+import compiler.utils.mutableKeyRefMapOf
+
+object GeneratorResolver {
+    fun computeGeneratorProperties(ast: Program): Map<Ref<Function>, List<Ref<Statement.ForeachLoop>>> {
+
+        val resultMapping: MutableMap<Ref<Function>, MutableList<Ref<Statement.ForeachLoop>>> = mutableKeyRefMapOf()
+
+        fun process(parentGenerators: List<Function>, statement: Statement) {
+            fun process(parentGenerators: List<Function>, vararg bunchOfBlocks: List<Statement>?) =
+                bunchOfBlocks.toList().forEach { block -> block?.forEach { process(parentGenerators, it) } }
```
ignacy123, line 16:
```suggestion
                bunchOfBlocks.forEach { block -> block?.forEach { process(parentGenerators, it) } }
```

app/src/main/kotlin/compiler/analysis/GeneratorResolver.kt
```diff
@@ -0,0 +1,62 @@
...
+import compiler.ast.Statement
+import compiler.utils.Ref
+import compiler.utils.mutableKeyRefMapOf
+
+object GeneratorResolver {
+    fun computeGeneratorProperties(ast: Program): Map<Ref<Function>, List<Ref<Statement.ForeachLoop>>> {
+
+        val resultMapping: MutableMap<Ref<Function>, MutableList<Ref<Statement.ForeachLoop>>> = mutableKeyRefMapOf()
+
+        fun process(parentGenerators: List<Function>, statement: Statement) {
```
ignacy123, line 14:
```suggestion
        fun process(surroundingGenerators: List<Function>, statement: Statement) {
```

app/src/main/kotlin/compiler/analysis/GeneratorResolver.kt
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        fun process(global: Program.Global) {
+            when (global) {
+                is Program.Global.VariableDefinition -> { }
+                is Program.Global.FunctionDefinition -> {
+                    val function = global.function
+                    val updatedParentGenerators = if (function.isGenerator) listOf(function) else listOf()
+                    if (function.isGenerator) resultMapping.putIfAbsent(Ref(function), mutableListOf())
+                    function.body.forEach { process(updatedParentGenerators, it) }
```
ignacy123, line 53: Maybe we can add `Function` type case to `process` and handle it exactly like that. Then both kinds of function definitions (global and not global) could act like simple pass-throughs.

app/src/test/kotlin/compiler/analysis/GeneratorResolverTest.kt
```diff
@@ -0,0 +1,250 @@
...
+import compiler.utils.keyRefMapOf
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class GeneratorResolverTest {
+
+    @Test fun `test generator with no foreach loop inside`() {
+        /*
+        Create AST for program:
+        ---------------------------------
```
ignacy123, line 19: I don't think this preamble is necessary

ignacy123 (changes requested):

app/src/main/kotlin/compiler/analysis/GeneratorResolver.kt
JHodorHodor: I did a little refactor to avoid the code repetition.

app/src/main/kotlin/compiler/analysis/GeneratorResolver.kt
```diff
@@ -0,0 +1,66 @@
...
+object GeneratorResolver {
+    fun computeGeneratorProperties(ast: Program): Map<Ref<Function>, List<Ref<Statement.ForeachLoop>>> {
+
+        val resultMapping: MutableMap<Ref<Function>, MutableList<Ref<Statement.ForeachLoop>>> = mutableKeyRefMapOf()
+
+        fun process(surroundingGenerators: List<Function>, node: AstNode) {
+            fun process(surroundingGenerators: List<Function>, vararg bunchOfBlocks: List<Statement>?) =
+                bunchOfBlocks.forEach { block -> block?.forEach { process(surroundingGenerators, it) } }
+
+            fun processFunction(surroundingGenerators: List<Function>, function: Function) {
```
ignacy123, line 21: Why not just inline this in line 38?

app/src/main/kotlin/compiler/analysis/GeneratorResolver.kt
```diff
@@ -0,0 +1,66 @@
...
+                val updatedSurroundingGenerators = if (function.isGenerator) surroundingGenerators + listOf(function) else surroundingGenerators
+                if (function.isGenerator) resultMapping.putIfAbsent(Ref(function), mutableListOf())
+                process(updatedSurroundingGenerators, function.body)
+            }
+
+            fun processForEachLoop(surroundingGenerators: List<Function>, foreachLoop: Statement.ForeachLoop) {
```
ignacy123, line 27: This should be inlined as well, we only call this once.

ignacy123 (approved):

app/src/main/kotlin/compiler/analysis/GeneratorResolver.kt
JHodorHodor: Because this is used twice

app/src/main/kotlin/compiler/analysis/GeneratorResolver.kt
JHodorHodor: Inlined

PR #163, opened 2023-01-14
Handle arrays in CFG generation

kamgal00:


app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
```diff
@@ -363,20 +405,94 @@ object ControlFlow {
...
+                is Expression.ArrayLength -> {
+                    val array = makeCFGForSubtree(astNode.expression)
+                    val resultReg = Register()
+                    cfgBuilder.addNextTree(
+                        IFTNode.RegisterWrite(
+                            resultReg,
+                            IFTNode.MemoryRead(IFTNode.Subtract(array, IFTNode.Const(memoryUnitSize.toLong())))
+                        )
+                    )
+                    cfgBuilder.addNextCFG(arrayMemoryManagement.genRefCountDecrement(array, expressionTypes[Ref(astNode.expression)]!!))
```
ignacy123, line 417: Where is the corresponding increment?

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
```diff
@@ -363,20 +405,94 @@ object ControlFlow {
...
+                    IFTNode.RegisterRead(resultReg)
+                }
+                is Expression.ArrayElement -> {
+                    val array = makeCFGForSubtree(astNode.expression)
+                    val index = makeCFGForSubtree(astNode.index)
+                    val elementAddress =
+                        IFTNode.Add(array, IFTNode.Multiply(index, IFTNode.Const(memoryUnitSize.toLong())))
```
ignacy123, line 424: No need to split the line here, I think :thinking:

stdlib/stdlib.c
```diff
@@ -27,6 +27,19 @@ void* _$checked_malloc(size_t size) {
     return address;
 }
 
+void _$populate_dynamic_array(uint64_t* address, uint64_t value, int64_t should_increment_elements) {
```
ignacy123, line 30:
```suggestion
void _$populate_dynamic_array(uint64_t* address, uint64_t value, int64_t should_increment_refcount) {
```

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
```diff
@@ -363,20 +405,94 @@ object ControlFlow {
...
-            cfgBuilder.addNextTree(variableAccessGenerators[Ref(owner)]!!.genWrite(targetVariable, result, owner === currentFunction))
-        } else if (accessNodeConsumer == null) {
-            cfgBuilder.addNextTree(result)
+        if (target != null) {
+            when (target) {
+                is AssignmentTarget.VariableTarget -> {
+                    val owner = variableProperties[Ref(target.variable)]!!.owner
+                    cfgBuilder.addNextTree(variableAccessGenerators[Ref(owner)]!!.genWrite(target.variable, result, owner === currentFunction))
+                }
+                is AssignmentTarget.ArrayElementTarget -> {
```
ignacy123, line 468: Why aren't we increasing the refcount here? The assigned variable will be a new reference to the array, right?

app/src/test/kotlin/compiler/intermediate/ExpressionControlFlowTest.kt
```diff
@@ -673,4 +690,83 @@ class ExpressionControlFlowTest {
                 )
             )
     }
+
+    private fun Int.toLiteral() = Expression.NumberLiteral(this.toLong())
+    private fun Int.toConst() = IFTNode.Const(this.toLong())
+
+    @Test
+    fun `simple array`() {
```
ignacy123, line 698: Some code snippets representing the tested operation would be nice (something similar to the tests above).

ignacy123 (changes requested):

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
kamgal00: It is done while evaluating `astNode.expression` to obtain `array` in the first line of the function.

app/src/main/kotlin/compiler/intermediate/ControlFlow.kt
kamgal00: The same reason as above

PR #164, opened 2023-01-15
Refactor the Compiler.process function

tomekb234:
This makes `Compiler.process` generally more readable. The interfaces for classes directly used by this function have been slightly changed, mostly by moving `diagnostics` parameter from the used function to its class' constructor. One notable change is that the vague class name `ControlFlow` has been changed to `ControlFlowPlanner` (because that's what this class basically does - it arranges the control flow of the program given its AST).

### DIFF
--- a/app/src/main/kotlin/compiler/Compiler.kt
+++ b/app/src/main/kotlin/compiler/Compiler.kt
@@ -2,11 +2,10 @@ package compiler
 
 import compiler.analysis.BuiltinFunctions
 import compiler.analysis.ProgramAnalyzer
-import compiler.ast.Function
 import compiler.ast.Type
 import compiler.diagnostics.Diagnostics
 import compiler.input.ReaderInput
-import compiler.intermediate.ControlFlow.createGraphForProgram
+import compiler.intermediate.ControlFlowPlanner
 import compiler.intermediate.FunctionDependenciesAnalyzer
 import compiler.lexer.Lexer
 import compiler.lowlevel.AsmFile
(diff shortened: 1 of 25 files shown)


app/src/main/kotlin/compiler/Compiler.kt
```diff
@@ -39,81 +38,101 @@ class Compiler(val diagnostics: Diagnostics) {
...
 
             val (functionDetailsGenerators, generatorDetailsGenerators) = FunctionDependenciesAnalyzer.createCallablesDetailsGenerators(
-                astWithBuiltinFunctions,
+                program,
                 programProperties.variableProperties,
                 programProperties.functionReturnedValueVariables,
-                diagnostics.hasAnyError()
-            )
+                diagnostics.hasAnyErrors()
+            ) // NOTE: it probably shouldn't be FunctionDependenciesAnalyzer responsible for this
```
ignacy123, line 71:
```suggestion
            ) // TODO: it probably shouldn't be FunctionDependenciesAnalyzer responsible for this
```

app/src/test/kotlin/compiler/intermediate/FunctionControlFlowPlannerTest.kt
```diff
@@ -113,21 +113,22 @@ class FunctionControlFlowTest {
...
     }
 
-    private fun test(program: Program) = ControlFlow.createGraphForEachFunction(
+    private val testDiagnostics = object : Diagnostics {
+        override fun report(diagnostic: Diagnostic) {
+            diagnostics.add(diagnostic)
+        }
+
+        override fun hasAnyErrors(): Boolean {
+            throw RuntimeException("This method shouldn't be called")
```
ignacy123, line 122: Maybe NotImplementedError instead?

test_all.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-./gradlew test "$@"
+./gradlew "$@" test install
```
ignacy123, line 2: Please update CI config to reflect this

ignacy123 (changes requested):

app/src/test/kotlin/compiler/intermediate/FunctionControlFlowPlannerTest.kt
tomekb234: I'm not the author of this code, but sure.

test_all.sh
tomekb234: Done.

ignacy123 (approved):

PR #165, opened 2023-01-15
Generators integration

tomekb234:
The base branch will be changed to `master` once #164 is merged.


app/src/main/kotlin/compiler/analysis/ProgramAnalyzer.kt
```diff
@@ -18,6 +19,7 @@ class ProgramAnalyzer(private val diagnostics: Diagnostics) {
         val defaultParameterMapping: Map<Ref<Function.Parameter>, Variable>,
         val functionReturnedValueVariables: Map<Ref<Function>, Variable>,
         val variableProperties: Map<Ref<AstNode>, VariablePropertiesAnalyzer.VariableProperties>,
+        val generatorProperties: Map<Ref<Function>, List<Ref<Statement.ForeachLoop>>>,
```
bebidek, line 22: `generatorProperties` is a very broad term while we have very specific mapping here.

tomekb234: I copied the interface from #162 and I agree that the name could be different.

tomekb234: Renamed.

ignacy123 (approved):

PR #171, opened 2023-01-17
Add AST node for array generation

tomekb234:
This seems to be all that's needed as the common interface for this week. Everything else will be mostly just syntax sugar on parse tree → AST level.

### DIFF
--- a/app/src/main/kotlin/compiler/ast/AstNode.kt
+++ b/app/src/main/kotlin/compiler/ast/AstNode.kt
@@ -30,6 +30,7 @@ sealed interface AstNode {
                         if (this.initializationType == Expression.ArrayAllocation.InitializationType.ONE_VALUE) arrayValues += "..."
                         "ciąg ${this.elementType}[${this.size.toSimpleString()}]{ $arrayValues }"
                     }
+                    is Expression.ArrayGeneration -> "ułożenie ${this.generatorCall.toSimpleString()}"
                 }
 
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/compiler/ast/AstNode.kt
```diff
@@ -30,6 +30,7 @@ sealed interface AstNode {
                         if (this.initializationType == Expression.ArrayAllocation.InitializationType.ONE_VALUE) arrayValues += "..."
                         "ciąg ${this.elementType}[${this.size.toSimpleString()}]{ $arrayValues }"
                     }
+                    is Expression.ArrayGeneration -> "ułożenie ${this.generatorCall.toSimpleString()}"
```
bebidek, line 33: This can lead to weird error messages when ciąg-comprehension (that appears in compiled code) will be shown as generator call (of some dummy generator), but I guess we can't do anything about it (at least without more serious changes in the project).

app/src/main/kotlin/compiler/ast/Expression.kt
```diff
@@ -47,6 +47,11 @@ sealed class Expression : AstNode {
         }
     }
 
+    data class ArrayGeneration(
+        val generatorCall: FunctionCall,
+        override val location: LocationRange? = null
+    ) : Expression()
+
```
bebidek, line 54: We had two options about "foreach in array" loop: syntax sugar or separate AST node. As far as I remember, we didn't settle on any option (so its up to the person who will implement that), so that person should be asked about that.

bebidek (approved):

PR #172, opened 2023-01-20
Array generation (phases after AST)

bebidek:

PR #175, opened 2023-01-20
Fix bugs in arrays

ignacy123:


tomekb234 (approved):
Changes for `stdlib.c` might need to be merged with #172.

bebidek:
If I understood it right, this PR changes semantics of `length` field in array: now it means total length of the structure (including two fields of metadata). Why do we change it? It is for some reason better this way?

stdlib/stdlib.c
```diff
@@ -31,10 +31,10 @@ void _$populate_dynamic_array(uint64_t* address, uint64_t value, int64_t should_
     if (address == 0)
         return;
 
-    uint64_t length = *(address - 1);
+    uint64_t length = *(address + 1);
```
bebidek, line 34: It seems to me that this method assumes that `address` parameter is given with different offset that the same parameter in `array_ref_count_decrement`, which is confusing.

ignacy123:
My intention was not to change it, in some places it was assumed that length is including the metadata, and in some not, which caused bugs. I just unified this to what I felt was intended.

stdlib/stdlib.c
ignacy123: I'll unify this

ignacy123:
I reverted the semantics of `length` to the old version in the most recent commit.

tomekb234 (approved):

app/src/main/kotlin/compiler/intermediate/generators/DefaultArrayMemoryManagement.kt
```diff
@@ -5,33 +5,40 @@ import compiler.ast.Type
...
-                    IFTNode.Add(mallocCall.result!!, IFTNode.Const(index * memoryUnitSize.toLong())),
+                    IFTNode.Add(IFTNode.RegisterRead(temporaryResultRegister), IFTNode.Const(index * memoryUnitSize.toLong())),
                     value
                 )
             )
         }
 
         writeAt(0L, IFTNode.Const(1L))
-        writeAt(1L, tableSize)
+        writeAt(1L, IFTNode.Subtract(totalTableSize, IFTNode.Const(2)))
```
bebidek, line 37: This looks little redundant as we still have `size` IFTNode.

app/src/main/kotlin/compiler/intermediate/ControlFlowPlanner.kt
```diff
@@ -485,15 +484,6 @@ class ControlFlowPlanner(private val diagnostics: Diagnostics) {
...
                     val accessGenerator = variableAccessGenerators[Ref(owner)]!!
-
-                    target.variable.type.takeIf { it is Type.Array }?.let {
-                        cfgBuilder.addNextCFG(
-                            arrayMemoryManagement.genRefCountDecrement(
-                                accessGenerator.genRead(target.variable, owner === currentFunction),
-                                it
-                            )
-                        )
-                    }
```
bebidek, line 496: Why is it removed? Is that case handled somewhere else?

app/src/main/kotlin/compiler/intermediate/ControlFlowPlanner.kt
```diff
@@ -517,12 +507,12 @@ class ControlFlowPlanner(private val diagnostics: Diagnostics) {
                 }
             }
         } else {
-            expressionTypes[Ref(expression)]?.let {
-                if (it is Type.Array) {
-                    cfgBuilder.addNextCFG(arrayMemoryManagement.genRefCountDecrement(result, it))
-                }
-            }
```
bebidek, line 524: Do I understand correctly that now if we provide `accessNodeConsumer` then calling function is responsible for adding the refcount decrement?
(I guess that is not going to happen anyway)

app/src/main/kotlin/compiler/intermediate/generators/DefaultArrayMemoryManagement.kt
bebidek: Never mind, I forgot there are two cases earlier.

app/src/main/kotlin/compiler/intermediate/ControlFlowPlanner.kt
ignacy123: I think we should first decrease (because the returning expression has just completed) and then increase (because the target variable is a new reference). I gathered from debugging that the increase is missing. To fix that, I just removed a decrease instead of adding the missing increase.

app/src/main/kotlin/compiler/intermediate/ControlFlowPlanner.kt
ignacy123: Do we want to decrease the reference counter if `accessNodeConsumer` is taking `it`? Doesn't that make the `accessNodeConsumer` a new owner of the current reference to `it`?

kamgal00: Yes, I think that `genRefCountDecrement` should be called if and only if `accessNodeConsumer` is null, otherwise `accessNodeConsumer` becomes new owner of that reference. I thought this case isn't possible anyway, but I forgot about `Statement.GeneratorYield`. In that case, I believe we don't need to worry about refcount - we can simply keep it incremented to the point when function that called generator receives it (we don't need to increment it there).

app/src/main/kotlin/compiler/intermediate/ControlFlowPlanner.kt
ignacy123: Nevermind, I was looking at the wrong bit of code

ignacy123: I didn't notice this removal is not necessary at all. Good catch :smile:

bebidek (approved):

PR #178, opened 2023-01-22
Introduce syntax of looping over arrays…

woocash2:
…as a syntax sugar of the regular loop.

Added e2e test for array loop.


app/src/main/kotlin/compiler/syntax/LanguageTokens.kt
```diff
@@ -39,6 +39,8 @@ object LanguageTokens {
             // Array related keywords
             Pair(TokenType.LENGTH, "długość"),
             Pair(TokenType.ARRAY_ALLOCATION, "ciąg"),
+            Pair(TokenType.LOOP, "dla"),
```
tomekb234, line 42: I think `ARRAY_FOR_EACH` would be more descriptive.

app/src/test/kotlin/compiler/e2e/CorrectProgramsE2eTest.kt
```diff
@@ -314,4 +314,18 @@ class CorrectProgramsE2eTest {
             """
         )
     }
+
+    @Test
+    fun `test looping over arrays`() {
```
tomekb234, line 319: Also at least one integration test (in the `tests/` folder) would be useful.

app/src/main/kotlin/compiler/syntax/AstFactory.kt
```diff
@@ -19,6 +19,8 @@ import compiler.syntax.utils.TokenRegexParser
 class AstFactory(private val diagnostics: Diagnostics) {
     class AstCreationFailed : CompilationFailed()
 
+    // We maintain the following counter to generate unique names of variables introduced by the associated syntactic sugar.
```
tomekb234, line 22: This comment seems to be outdated.

tomekb234 (changes requested):

app/src/main/kotlin/compiler/syntax/AstFactory.kt
```diff
@@ -409,6 +411,60 @@ class AstFactory(private val diagnostics: Diagnostics) {
                 val bodyBlock = processMaybeBlock(children[4])
                 Statement.Loop(conditionExpr, bodyBlock, combineLocations(children))
             }
+            in listOf(Productions.nonBraceStatementArrayLoop, Productions.nonIfNonBraceStatementArrayLoop) -> {
+                val arrayExpression = processExpression(children[6])
```
tomekb234, line 415: Please extract this code into a method (e.g. `makeArrayLoop`) which could also be used in #179.

app/src/test/kotlin/compiler/e2e/CorrectProgramsE2eTest.kt
woocash2: You mean in `AstFactoryTest` for example?

tomekb234: A test to check if a correct AST is produced (in `AstFactoryTest`) would also be welcome, but what I mean is a simple integration test which checks that the loop actually works when running the program (in the `tests/` folder in the project root).

woocash2: Will add tomorrow, after I decode how to do it.

extras/vim/syntax/brzeczyk.vim
```diff
@@ -23,7 +23,7 @@ syn keyword nnKeywordBoolLiteral fałsz prawda
 syn keyword nnKeywordOtherLiteral nic
 syn keyword nnKeywordCondition jeśli wpp
 syn match   nnKeywordElif '\<zaś gdy\>'
-syn keyword nnKeywordLoop dopóki otrzymując od
+syn keyword nnKeywordLoop dopóki otrzymując od dla
```
JHodorHodor, line 26: `wewnątrz`?

woocash2: Yeah

app/src/main/kotlin/compiler/syntax/AstFactory.kt
```diff
@@ -106,6 +106,63 @@ class AstFactory(private val diagnostics: Diagnostics) {
         return rotateExpressionLeft(newTopLevelNode)
     }
 
+    fun makeArrayLoop(parseTree: ParseTree<Symbol>): Statement {
```
tomekb234, line 109: I meant for this function to take AST nodes as parameters.

JHodorHodor: I think that this is also fine. I will have to refactor it a little bit merging anyway.

tomekb234 (approved):

app/src/main/kotlin/compiler/syntax/AstFactory.kt
woocash2: Which ast nodes? `arrayExpression`, `receivingVariable` and `bodyBlock`?

woocash2: let's go

